PR #130, opened 2015-08-24
moved ExpressionParser from AbstractTestClass to its own class to remove dependencies on stext test plugin

bohl-itemis:
dependencies to test packages.


andreasmuelder:
Merged into master

PR #281, opened 2016-01-06
added sct-unit and sct-model for issue 278 ( No action executed on

DominikTesch:
outgoing transitions from an initial state )


terfloth:
We have to create a new pull request for that issue branch

PR #407, opened 2016-02-17
Unresolveable type proxies for when a snapshot is restored

andreasmuelder:


andreasmuelder:
@tkutz review please - does this fix also work for other domains than generic?

PR #426, opened 2016-02-18
Issue 412

andreasmuelder:
Filter the Validators for a given Domain to avoid checks with different type systems


andreasmuelder:
@tkutz could you review this PR ?

tkutz:
works for me

PR #446, opened 2016-02-22
User documentation: PDF generation deactivated by default

RainerKlute:


andreasmuelder:
Closed sa duplicate with pr #471

Issue #694, opened 2016-05-19
Types need proper ID attributes


tkutz:
In the type meta model, elements do not have any id attribute. This results in some flaws when referencing such type elements in another model, since proxies will have indices based on the type position. As these might change, the references get broken.

tkutz:
Fixed with pull request #695.

PR #695, opened 2016-05-19
Proper ID attributes in type meta model (#694)

tkutz:
PackageMember has now "id" attribute which is an EMF ID.
PackageMemberImpl implements default strategy to derive ID:
* iterate over all containers and if they have a name, prepend it with "." as separator.

The "getId()" methods needs to be re-implemented in DeclarationImpl, TypeAliasImpl and LocalReactionImpl due to multiple inheritance restriction (not inheriting from super impl class).


tkutz:
@andreasmuelder please review or let's discuss on Monday.

tkutz:
ID attribute for PackageMember is currently not possible because LocalReaction inherits from it. Therefore I changed it once more to only provide ID attribute in Operation, Property and Type.

Issue #610, opened 2016-04-09
Changes to sc_types.h are overwritten


stmork:
In past versions of SCT the file sc_types.h had a special treatment so that the types can be adjusted to the platform needs. Now the changes are lost after regenerating the C++ artefacts.

nyssen:
This can now ne achieved by setting a libraryTargetFolder within the Outlet feature of the sgen file. All static files, i.e. those not depending on the concrete model (including sc_types.h) will be generated into it, and will be preserved upon re-generation. This way, they can now not only be manually adopted but also be shared by different statecharts in one project.

stmork:
This simply does not work and the intended behaviour is undocumented.

stmork:
I have tried following variants without success:
1. setting `libraryTargetFolder` equal to `targetFolder`
2. setting `libraryTargetFolder` to "."
3. setting `libraryTargetFolder` to a directory different to `targetFolder`

jdicks:
pull request #703

jdicks:
fixed by https://github.com/Yakindu/statecharts/pull/703

PR #703, opened 2016-05-23
#610 : CPPgenerator avoids overriding of sc_types.h

jdicks:


jdicks:
@terfloth : plz review

PR #742, opened 2016-06-04
Fix installation instructions link

hudsonsferreira:


tkutz:
Thanks for spotting this!

PR #762, opened 2016-06-16
Documentation: changes to generate online documentation

RainerKlute:
* CSS modified.
* Textile files must start with an empty line, because they are concatenated.
* "Blurred" version of two image files provided. Originals are available in XCF (not deployed).
* Textiles source files must still be deployed until PRO and OSS documentation is unified in the OSS repository.


tkutz:
please verify on all common browsers if styling still works

PR #785, opened 2016-06-24
Create GMFResource if default/gernic domain does not exist and resour…

jdicks:
…ce is not existing in filesystem


jdicks:
fix for #765

Issue #793, opened 2016-06-27
Remove 'examples/traffic_light_arduino'


jdicks:
as .elf is a binary format it isn't well compareable and mostly marked as changed when using git/eGit.

Further the whole folder 'traffic_light_arduino\Release' is a majorly platform dependant result and therefore should not be present in the repository.

andreasmuelder:
@jdicks can you file this in the new Yakindu/examples repository, all examples contained inYakindu/statecharts will be deleted soon.

jdicks:
see https://github.com/Yakindu/examples/issues/1

jdicks:
@BeckmaR will take over the issue in example repository. i will delete the sample from statechart repository.

PR #795, opened 2016-06-27
#793 : remove C application

jdicks:


andreasmuelder:
Can you apply those code changes to the example in https://github.com/Yakindu/examples
The examples in Yakindu/statecharts will be deleted soon.

jdicks:
removed the whole sample because it is already moved to https://github.com/Yakindu/examples/tree/master/traffic_light_arduino

Issue #854, opened 2016-07-26
Transition priority change propagation of vertices


jdicks:
For https://github.com/Yakindu/statecharts/issues/791 a general synchronization between the priority property and the priority decorator was introduced.

this is working for states as expected but not working for choices and synchronizations.

the proper feature for change notifications is vertex_outgoingtransitions instead of state_outgoingtransitions

andreasmuelder:
@jdicks is this fixed?

PR #855, opened 2016-07-26
#854 : notify changes of vertex instead of state, (& defense casting)

jdicks:


plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/providers/TransitionPriorityDecorationProvider.java
```diff
@@ -97,8 +97,8 @@ public void notifyChanged(Notification notification) {
 		@Override
 		public void activate() {
 			super.activate();
-			if (semanticElement instanceof Transition
-					&& ((Transition) semanticElement).eContainer() instanceof Vertex) {
+			if (!(semanticElement instanceof Transition)
+					&& !(((Transition) semanticElement).eContainer() instanceof Vertex)) {
```
andreasmuelder: This should ne an || instead of a &&

Issue #384, opened 2016-02-10
Missing syntax docu for transition


DominikTesch:
A detailed docu for transition synatx is missing.

RainerKlute:
Fixed with pull request #902.

PR #902, opened 2016-08-15
Transitions documented (issue #384)

RainerKlute:
@andreasmuelder, please proofread!


plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
 
 h3(#transition). Transition
 
-A transition is the transfer of one state to another. Transitions are diagrammed as arrows. They can have events and actions, but don't need to.
+A transition is the transfer of one state to another. A transition is diagrammed as a arrow leading from a _source state_ to a _target state_. A transition can also have
```
tkutz: as *an* arrow

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+"Figure &quot;Transition syntax overview&quot;":#fig_transition-syntax-overview shows how the syntax of a transition specification is defined. You can refer to this diagram while reading the subsequent sections.
+
+p(#fig_transition-syntax-overview). 
+!(standard-image)images/docu_grammar_030_transition-overview.png(Transition syntax overview)!
+
+p=. Transition syntax overview
+
+h4(#unspecified-transition). Unspecified transition
+
+In the most simple case we have only the transition's arrow leading from a source state to a target state without any additional specification. In other words, there is no text attached to the transition. However, such an event will never be taken, so better make sure to set up a transition specification.
```
tkutz: That is not true. When nothing is specified, (1) we get a warning and (2) that transition is always taken.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+For a transition to fire, the triggering event or events must be given in the transition specification as an _event trigger_. An event trigger consists of one or more event specifications, separated by comma. An event specification is either the name of an event that has been declared in the definition section, see "section &quot;Events&quot;":#events, or one of the built-in event specifications _after_, _every_, _always_, _default_, _else_, or _oncycle_. The latter are documented in "section &quot;Reaction triggers&quot;":#reaction-triggers.
+
+Let's have a look at some sample transition specifications. We are assuming that the transition's source state is active, because otherwise the transition wouldn't fire anyway.
+
+| *Event trigger:* | *Prerequisites for the transition to fire:* |
+| @ev@ | Event _ev_ occured. |
+| @ev_1, ev_2@ | At least one the events _ev_1_ and _ev_2_ occured. |
+| @after 3s@ | Three seconds after activating the source state have passed. |
+| @after 3s, ev@ | Three seconds have passed or event _ev_ occurred, whatever happens first. |
+| @after 3s, after 2s@ | Three seconds have passed or two seconds have passed, whatever happens first. :-) |
```
tkutz: This is redundant to 2nd example

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+
+For a transition to fire, the triggering event or events must be given in the transition specification as an _event trigger_. An event trigger consists of one or more event specifications, separated by comma. An event specification is either the name of an event that has been declared in the definition section, see "section &quot;Events&quot;":#events, or one of the built-in event specifications _after_, _every_, _always_, _default_, _else_, or _oncycle_. The latter are documented in "section &quot;Reaction triggers&quot;":#reaction-triggers.
+
+Let's have a look at some sample transition specifications. We are assuming that the transition's source state is active, because otherwise the transition wouldn't fire anyway.
+
+| *Event trigger:* | *Prerequisites for the transition to fire:* |
+| @ev@ | Event _ev_ occured. |
+| @ev_1, ev_2@ | At least one the events _ev_1_ and _ev_2_ occured. |
+| @after 3s@ | Three seconds after activating the source state have passed. |
+| @after 3s, ev@ | Three seconds have passed or event _ev_ occurred, whatever happens first. |
```
tkutz: This is redundant to 2nd example

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+| @after 3s, after 2s@ | Three seconds have passed or two seconds have passed, whatever happens first. :-) |
+| @every 3s@ | Three seconds after activating the source state have passed. |
```
tkutz: This event triggers every 3 seconds, in contrast to *after* which only triggers once after 3s. This is relevant in combination with a guard: if the condition is not fulfilled after 3s it will be checked again after 3s, whereby with *after* it will be only checked once after activating the source state

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+| @always@ | Transition fires always. |
+| @default@ | Transition fires always. |
+| @else@ | Transition fires always. |
```
tkutz: *default* and *else* are basically the same and only meaningful with choices to define a transition to be taken when all other transitions of the choice cannot be taken.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+| @oncycle@ | Transition fires always. |
```
tkutz: *oncycle* and *always* are basically the same. I would group them together.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+
+h4(#guard-condition). Guard condition
+
+A _guard_ applies an additional condition for a transition to fire. It is not sufficient for the specified event to occur, the guard condition must also be fulfilled. It is written in square brackets following the event trigger. It is also possible to specify a guard without any event trigger. In this case the transition is always taken, as long as the guard condition evaluates to _true_.
+
+The following examples assume the _x_ is an integer variable and _isOpen_ is an operation.
+
+| *Transition specification:* | *Prerequisites for the transition to fire:* |
+| @ev [x == 3]@ | Event _ev_ occured and _x_ holds the value 3. |
+| @ev_1, ev_2 [x == 3]@ | At least one the events _ev_1_ and _ev_2_ occured and _x_ holds the value 3. |
```
tkutz: at least one *of* the events...

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -46,11 +46,120 @@ States are the central elements of a state machine. A state has to be placed ins
...
+It is not possible to have event-specific guards as in the illegal @ev_1 [x == 1], ev_2 [x == 2]@. If you need something like this, use multiple transitions with @ev_1 [x == 1]@ on one of them and @ev_2 [x == 2]@ on the other.
+
+h4(#effect). Effect
+
+A transition specification may contain one or more _actions_ that are executed when the transition fires, like assigning a value to a variable, calling an operation, or raising an event. All actions taken together are called the transition's _effect_.
+
+In the transition specification, the effect is written right from event trigger and guard and separated by a @/@ character from them. "Figure &quot;Transition effect syntax&quot;":#fig_transition-effect-syntax shows how the effect's syntax looks like.
+
+p(#fig_transition-effect-syntax). 
+!(standard-image)images/docu_grammar_030_transition-effect.png(Transition effect syntax)!
```
tkutz: It is a technical detail that we distinguish between expression and event raising expression. In the figure it should be only one; and then the figure does not tell you much anymore; I think it can be omitted.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -627,3 +738,52 @@ bc. myBool = active(StateA)
...
+
+h3(#expressions-grammar). Expressions grammar
+
+Figure "&quot;Expressions grammar&quot;":#fig-expressions-grammar shows the expressions grammar. It is used by the "statechart grammar":#statechart-grammar.
+
+p(#fig-expressions-grammar). 
+
+!(standard-image)images/docu_grammar_010_expressions.png(Expressions grammar)!
+
+p=. Expressions grammar
```
tkutz: I don't think the complete language grammar is really useful. The users who are really interested in its details because they want to adjust the language, would need to read the xtext grammar file.

I think it is best to extract some interesting parts of it and use it in the documentation (transition, state, statechart specification) instead of having the whole grammar on one shot.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: Okay, I'll remove the `after 3s, after 2s` example, although I consider it a little funny.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: Actually it isn't redundand, because it show that you can mix events and timings.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: Agreed on Expression vs. EventRaisingExpression, however, until we have a railroad chart editor, we'll have to deal with what we have.

The chart visualizes that expressions are separated by semicolon; so I think I'll keep it.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: I think it doesn't harm to keep it here. Nobody is _required_ to read it, but in case someone is interested in the complete grammar – and if just out of curiosity –, it is more convenient to browse it here than to download the source code distribution and find out a) _where_ the grammar is defined, b) _how_ the grammar is defined, c) what that Xtext is all about, and d) how to understand an Xtext file. Providing the railroad diagrams lowers the hurdles.

tkutz:
Closing, because #911 contains these changes.

PR #941, opened 2016-09-05
Introduced key word 'state machine' for SEO

svenjawendler:


RainerKlute:
I'd rather reject that change, because the topic here is to *start working with SCT*, not *modeling state machines*. Using "state machine" is recommendable, but we don't have to introduce it everywhere no matter the cost.

Issue #35, opened 2015-04-24
Handle "open file" gracefully 


GoogleCodeExporter:
```
The SCT editor does not support opening act-files from the file system. The 
editor crashes if this is done.
```

Original issue reported on code.google.com by `AxelTerfloth` on 29 Sep 2014 at 1:52

BeckmaR:
Fixed with PR #968

BeckmaR:
Pull request merged, closing.

PR #969, opened 2016-09-09
Added null tests & error message for file -> open file usage #35

BeckmaR:
Fixes the weird issue when a user attempts to open a statechart via file -> open file. Only an empty editor was displayed, but no error message. An error message was added and the number of exceptions reduced to a minimum through a number of null-checks in YAKINDU classes (issue #35)


BeckmaR:
New PR for #968, where EGit produced a weird diff

Issue #574, opened 2016-04-04
Validation marker inconsistencies


jdicks:
If i add a transition and state to the statechart via the semantic element proposals (using strg+space > "add outgoing state"), validation issues will not be shown/will be removed in the graphical editor after saving.

after inserting there should be 2 errors like:
- A state must have a name. default.sct /test.184   line: 0 /test/default.sct   Statechart Diagram problems
- Missing trigger. Transition is never taken. Use 'oncycle' or 'always' instead default.sct /test.184   line: 0 /test/default.sct   Statechart Diagram problems

The markers are shown in the problems view and other views (e.g. Project Explorer) but never be shown in the SCT editor itself.

When the state or transition is touched (e.g. moved) and the model is saved again the markers are shown as expected.

The problem might be that the sematic element(s) aren't associated properly with a gmf element, so  the marker can't be shown in the editor.

DominikTesch:
I add region and an entry point, 2 states (with name) and two transitions via the tools list. If I typ "# x = true" as trigger for the Transition between the states an error mark appears and if I change it to "[x == true]" I will get this errors.
Transitions with an exit point spec does not have a trigger or guard.

When you save the error will disappear but when you change something the error will come back.

jdicks:
In another model i observed that 2 types of error markers are shown in problems log.

![markers](https://cloud.githubusercontent.com/assets/5557073/16809626/4a6f0d6e-4921-11e6-941d-a38140770d7c.PNG)

jdicks:
this includes 2 bugs resulting from
1. transitions properties aren't cleared before the will be parsed again
2. the semantic proposal 'Add outgoing state' didn't add notation elements so that error markers could not be associated to view elements

 see pr #1028

jdicks:
was reverted before release 2.8.0 because inserting elements via the semantic proposal leads into overlapping elements. This is caused by absolute bounds set for the notation elements during insert (see PR #1028)

jdicks:
extracted #1054
* removing transition properties before parsing again
** solves the inconsistencies mentioned by  DominikTesch (on 5 Apr)
** :exclamation: because it can be applied without further changes

jdicks:
To fix inconsistencies mentioned from jdicks on 13 Jul &  on 4 Apr follwoing aspects are important:

* ensure notation elements are serialized even if they are derived during GEF/GMF initialization
[...]

PR #1028, opened 2016-10-12
#574 : fix error marker inconsistencies

jdicks:
… AddOutgoingStateModification creates notation elements

### DIFF
--- a/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
+++ b/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
@@ -10,17 +10,13 @@
  */
 package org.yakindu.sct.ui.editor.editor.proposals;
 
-import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.gmf.runtime.diagram.core.services.ViewService;
 import org.eclipse.gmf.runtime.notation.Bounds;
-import org.eclipse.gmf.runtime.notation.Diagram;
 import org.eclipse.gmf.runtime.notation.LayoutConstraint;
 import org.eclipse.gmf.runtime.notation.Node;
 import org.eclipse.gmf.runtime.notation.NotationFactory;
 import org.eclipse.gmf.runtime.notation.View;
-import org.eclipse.xtext.EcoreUtil2;
-import org.yakindu.sct.model.sgraph.Region;
 import org.yakindu.sct.model.sgraph.SGraphFactory;
 import org.yakindu.sct.model.sgraph.State;
 import org.yakindu.sct.model.sgraph.Transition;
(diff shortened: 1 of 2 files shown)


plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
```diff
@@ -67,15 +78,63 @@ protected boolean check(EObject semanticObject, View view) {
...
+		Bounds sourceBounds = null;
+		if(sourceView instanceof Node){
+			Node mynode = (Node)sourceView;
+			LayoutConstraint layoutConstraint = mynode.getLayoutConstraint();
+			if(layoutConstraint instanceof Bounds){
+				sourceBounds = (Bounds) layoutConstraint;
+			}
+		}
+		Bounds newStateBounds = NotationFactory.eINSTANCE.createBounds();
+		newStateBounds.setX(sourceBounds.getX()+200);
```
jdicks: Any other idea? i did not find the place where the location was set before and decided to create a state 200 points right to the source state statically.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
```diff
@@ -67,15 +78,63 @@ protected boolean check(EObject semanticObject, View view) {
...
+		newStateBounds.setY(sourceBounds.getY());
+		stateNode.setLayoutConstraint(newStateBounds);
+		return stateNode;
+	}
+	/**
+	 * @see org.yakindu.sct.ui.editor.factories.FactoryUtils.getRegionCompartmentView(View)
+	 */
```
andreasmuelder: Why @see FactoryUtils? why can't we use this implementation here?
return (View) regionView.getChildren().get(1);

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
andreasmuelder: See Ctor of CreateViewAndOptionallyElementCommand

jdicks: Does not help, because CreateViewAndOptionallyelementCommand works with draw2d EditParts instead of Notation elements.
i just figured out that the element was added ~ source.x + 150 & source.y + 50 and added it hard coded into the bounds calculation of the new state view.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/proposals/AddOutgoingStateModification.java
jdicks: Because the given view (from source state) is a child of the needed region compartment view, end up with the following:
	private View getRegionCompartmentView(View stateView) {
		return (View) stateView.eContainer();
	}

PR #1040, opened 2016-10-17
Enabled context based filtering for type references.

tkutz:


jdicks:
@andreasmuelder : plz review

PR #1088, opened 2016-11-16
Improve meta model

BeckmaR:
Improve the meta model by introducing more subclasses, allowing us to remove many dispatch-functions from the code generators.

Don't merge yet, this is a Travis-test-playground.


BeckmaR:
@andreasmuelder This can be merged now.

PR #1096, opened 2016-11-21
Clean up documentation

BeckmaR:
Some chapters were swapped in order to make more sense logically. This also
allowed to remove some sub-sections, because they then appeared in chapters
directly following each other.

Some typos were fixed, and HTML excaped chars were removed, because they
are actually escaped automatically.

The chapter C++ features was moved to the C generator, because it appears
before the C++ one and linked to it. The feature exclusive to the C++ generator
stayed there.

The Java runtime got a mention of the respective Feature parameter.


BeckmaR:
@RainerKlute maybe you want to take a look at this?

PR #1124, opened 2016-12-06
Corrected misspelled basePackeName static return value

lionax:
Just a little fix in the java code generator I came across


andreasmuelder:
@lionax thanks for contributing! I think the .scr was intentional as 'statechart runtime'. However, I think that org.yakindu.sct is more intuitive.  I think we can merge this, @terfloth what do you think?

terfloth:
We can't merge this as this will break the established default. 👎

PR #1126, opened 2016-12-08
Issue 1109

terfloth:


plugins/org.yakindu.sct.generator.cpp/src/org/yakindu/sct/generator/cpp/Types.xtend
```diff
@@ -28,14 +28,18 @@ class Types implements IContentTemplate {
 		#ifndef «typesModule.define»_H_
 		#define «typesModule.define»_H_
 		
-		#define sc_string		char*
+		//#include <cstdint>
```
jdicks: Is there a particular reason to generate this commented include?

terfloth: Yes the disabled include is intended as it is an alternative that can be chosen by the user. Added comment to generated code to make that clear ...

PR #1127, opened 2016-12-08
New testcase for headless code generation

DominikTesch:


andreasmuelder:
@jdicks @DominikTesch This is something that should be tested with an automated unit test instead of manual testing in the testplan

jdicks:
@andreasmuelder from my point of view the manual test case is a good point to start even if we might implement a junit test soon. Further the installation aspect will be hard to cover by a junit test so it might be a good idea to keep a manual test case for it.

PR #1138, opened 2016-12-15
adjust headless user guide (single model parameter)

jdicks:


jdicks:
@RainerKlute or @terfloth plz review

Issue #1174, opened 2017-01-23
Example wizard doesn't display examples from downloaded example repo


BeckmaR:
After downloading the zip from github and specifying it in the settings, its contents are not displayed.

PR #1179, opened 2017-01-24
Check for existence of metadata.json files instead of .git folder (#1…

tkutz:
…174)

Checking for .git folder only works for cloned repositories, but using downloaded zip from github does not work as it does not contain the .git metadata folder.


tkutz:
Fixes #1174

Issue #1194, opened 2017-02-09
Improve cancelation of SCTValidationJob


jdicks:
# Description:
In SCT the editor use a the SCTValidationJob to validate the resource content.
The job is scheduled by the ResourceSetValidationListener which is a ResourceListener which is notified after each execution of a transaction (see EMF transaction).

Before new scheduling of the job, a potential already running job is cancelled. This should ensure the liveness and if the user do fast interactions with the editor.

This implementation seems to work fine as long as you work with small models. But it actually does not work.
This gets obvious if you work with large models, where the validation job runs longer than some milliseconds.
Than the validation job actually blocks the user interaction.

# Root cause:
The location to cancel a running job is wrong.
The job should be cancelled after a new notification indicates a new subsequent change of the model. But because of the locking of the editing domain during the validation it is not possible to create a new change which change the model and leads to such a new notification.

# Possible solution:
It is necessary to ensure, that the running job is actually cancelled before a user interaction leads to asking for a new lock on the editing domain.
For this purpose the only location which come into my mind is the TransactionalCommandStack. This is a central location where each user interaction is going through. So in the methods execute, undo and redo, a cancel and reschedule of the job will solve the problem.

# Additional Info: 
I found out, that in GMF use case the canceling of the validation job on executing a command in the command stack like described in this bug is not sufficient.
GMF very very often calls org.eclipse.emf.transaction.TransactionalEditingDomain.runExclusive(Runnable) which also blocks the ui thread until read only access to the model could be granted.
So also on this case the validation job must be canceled. Anyway it is much more tricky to get a good UX here, because this canceling happens really very often in GMF. So it could be easily the case, that a validation which really takes long may never ends. Also validation may only be rescheduled if really was canceled beforhand...

jdicks:
attached a prototypical patch the StatechartDiagramEditor:
[StateMachineEditor.txt](https://github.com/Yakindu/statecharts/files/764118/StateMachineEditor.txt)

andreasmuelder:
I propose to not execute the Validation Job within a runexclusive on the TransactionalEditingDomain, see #1249 The validation job does not block UI interaction and allows to change the semantic model during validation. If this happens, the Job gets canceled and restarted.

andreasmuelder:
Second approach committed to branch is a ShadowModelValidation Job that clones the whole resource before validation und validates on the shadow model without a read transaction. More time and memory consuming but way more responsive  UI.

andreasmuelder:
[...]

PR #1201, opened 2017-02-17
Issue 1190

andreasmuelder:


plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/StatechartDiagramEditor.java
```diff
@@ -151,9 +145,13 @@ public void init(IEditorSite site, IEditorInput input) throws PartInitException
 	}
 
 	private void registerValidationListener() {
```
maximilianost: Might be helpful to make it protected, so that clients can override.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/SCTValidationJob.java
```diff
@@ -120,45 +111,23 @@ public boolean isCanceled() {
 			try {
 				editingDomain.runExclusive(runner);
```
maximilianost: Please make this overrideable by extracting to a protected method.
In my case, I have to run the validation e.g. in a write transaction and rollback afterwards.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/SCTValidationJob.java
```diff
@@ -40,20 +34,18 @@
...
  * 
  * @author andreas muelder - Initial contribution and API
  * 
  */
-@Singleton
-public class SCTValidationJob extends Job implements IMarkerType {
+public class SCTValidationJob extends Job {
 
 	@Inject
 	private IResourceValidator validator;
```
maximilianost: By running the validation with the IResourceValidator you get the overhead of issue creation and this results in your API design for IValidationIssueProcessor.
I suggest not to work with issues at all. Just use the Diagnositcian for validation and work completely with Diagnostic.
You will get FeatureBasedDiagnositcs which will be more easily be connectable to the target EObjects.

But this is only a personal preference of mine.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/DefaultValidationIssueStore.java
```diff
@@ -0,0 +1,207 @@
...
+		connected = true;
+		this.resource = resource;
+		reloadMarkerIssues();
+	}
+
+	protected void reloadMarkerIssues() {
+		persistentIssues.clear();
+		if (liveValidationEnabled())
+			return;
+		IFile file = WorkspaceSynchronizer.getFile(resource);
```
maximilianost: Not sure why you connect this store implementation to the persistent marker world.
In the SctEditor you schedule a validation right after initialization. So the live markers here will be updated and are (hopefully) in sync with the persistent markers.
So this seems to me as a unecessary feature. I think it is better not to merge the use cases of live in editor validation (and its issues/diagnostics) and the persistent marker world.

And if you do that this way, I think that at least the initial scheduling of the validation is superfluous.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/SCTValidationJob.java
andreasmuelder: This will ne part of https://github.com/Yakindu/statecharts/issues/1194

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/DefaultValidationIssueStore.java
andreasmuelder: In SCT there is the option to disable live validation in the preferences. If you do that, you still want to see the persistent error markers in your editor. Thus, the connection to the marker store.

This is also preliminary work for stale error markers. These are error markers that are present as a marker (unsafed changes) but already fixed in the editor. In JDT and Xtext editors, they are visualized as grayscale error icons.  To calculate stale error markers I need the IMarker based issues aswell.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/SCTValidationJob.java
andreasmuelder: We use Issues here to reuse parts of the Xtext infrastructure (IValidationIssueProcessor, MarkerCreator, MarkerTypeProvider, etc.) The overhead of Issue creation can be ignored since these are just simple POJOs. If it pops up in your profiler please let me know.

andreasmuelder:
@maximilianost thanks for the comments, @jdicks @tkutz any other review comments on this issue?

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/StatechartDiagramEditor.java
andreasmuelder: done

Issue #684, opened 2016-05-11
Extract Installation chapter online documentation


RainerKlute:
The following originated as issue #663, which has been split into two different issues:

The current status of the online documentation is that the user needs to read and scroll a lot to get a first impression of the tool. It basically starts with a long installation part before giving a first impression on how the tool looks like. The installation part gives the impression that it must be complicated to install the software as it contains 2229 words and 33 screenshots to express that you just need to unzip the downloaded file on each platform.

Especially with the need to give some personal data (email, name, etc.) before downloading the tool, we should give the potential user a good first impression on how the tool looks like when he visits the online documentation.

We should do the following:
* Our current Yakindu Screencast is not part of video section (only the Tom Mens video), it should be also added there.
* Move installation part to the end of online documentation.

RainerKlute:
See https://www.itemis.com/en/yakindu/statechart-tools/tutorials/video-tutorials/ for the reworked video tutorial page.

PR #1227, opened 2017-02-23
#650 - Adopt meta model to support generic operations

andreasmuelder:


BeckmaR:
If this is merged, pro is broken until its PR #684 is merged too.

PR #1260, opened 2017-03-09
Template type inferrer

BeckmaR:
TypeInferrer improved to correctly infer type / typearguments of generic elements and general infrastructure improvements.


BeckmaR:
From my point of view, we could merge this now. @tkutz?

PR #1296, opened 2017-03-28
trivial : export refactoring handler implementations

jdicks:
gives the possibility to reuse concrete refactoring handler
implementations within derived editors.


andreasmuelder:
@jdicks is this still required?

PR #1302, opened 2017-04-03
Tree naming service preparations

BeckmaR:
The Naming service was split up, so that all language specific functions are not anymore part of the Naming Service.


BeckmaR:
Work suspended for now.

PR #1306, opened 2017-04-04
 Maintain visible issues from persitent markers & live issues

jdicks:
* avoid union of live validation issues & persistent markers
  * avoids markers are shown which are already fixed in persistent state
* use persistent state for visible issues once the resource changed
(marker change)
  * reexecution of live validation is not needed
* maintain visible issues based on live issues after live validation is finsihed
  * this hides persitent markers which are fixed in dirty state


jdicks:
@andreasmuelder: plz have a close look to synchronization of visible markers within DefaultValidationIssueStore

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/validation/DefaultMarkerChangeProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
+import com.google.inject.Inject;
+
+/**
+ * Encapsulates functionalities to process resource deltas markers and convert them into an efficient data structure.
+ */
+public class DefaultMarkerChangeProcessor implements IMarkerChangeProcessor {
```
jdicks: think about a stateless implementation (ny @andreasmuelder )

PR #1335, opened 2017-05-03
Enforce minimum height of documentation property section

Bananeweizen:
This change sets the minimum height of the documentation text field in
the properties to be 3 times the height of a single line. In practice at
least on my windows machine still only 2 lines (and some empty pixels)
are visible at any time.

If this is not done, the text control resizes up to no height at all, if
the remaining controls (like list of transitions in the state) gets
long.

The main change is the line containing "minSize", everything else is removal of duplicated code.

### DIFF
--- a/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/propertysheets/AbstractTwoColumnEditorPropertySection.java
+++ b/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/propertysheets/AbstractTwoColumnEditorPropertySection.java
@@ -10,11 +10,14 @@
  */
 package org.yakindu.sct.ui.editor.propertysheets;
 
+import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.SashForm;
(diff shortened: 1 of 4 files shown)


andreasmuelder:
@BeckmaR Could you check this for Linux and @tkutz could you check if this is working on Mac?

tkutz:
Works on Mac, but only shows 2 and a half lines instead of 3.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/propertysheets/AbstractTwoColumnEditorPropertySection.java
```diff
@@ -57,4 +62,11 @@ protected Layout createRightColumnLayout() {
 		return createBodyLayout();
 	}
```
andreasmuelder: I am not sure if the createDocumentationControl should be part of the AbstractTwoColumnEditorPropertySection. Not every TwoColumnEditorPropertySection needs a DocumentationControl.  Previous version however had duplicate code so this is fine for me.

andreasmuelder:
@Bananeweizen thanks for your contribution, merged into master

PR #1352, opened 2017-05-05
fix broken markup in readme

Bananeweizen:
Github tries to render the text "<User>" as tag, so escape it.


andreasmuelder:
thanks!

PR #1353, opened 2017-05-08
Changed the method validateParameter from protected to public

jthoene:
The method is needed in the SCTUnitValidator so it needs to be public.


andreasmuelder:
This can be handled in SCTUnitTypeInferrer

PR #1365, opened 2017-05-10
Issue 1336

### LINKED ISSUE #1336: Avoid magic numbers in C code generator
There are plenty of magic numbers generated that should rather be defines to avoid QAC warnings:
image002
image003

BeckmaR:
Resolves #1336 by calculating and applying defines for all states that can be used in the code instead of "magic" indices of the StateConfVector.

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StateConfVectorIndexCalculator.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StateConfVectorIndexCalculator.xtend
@@ -1,11 +1,15 @@
 package org.yakindu.sct.generator.c
 
+import com.google.inject.Inject
 import java.util.HashMap
 import java.util.Map
 import org.yakindu.sct.model.sexec.ExecutionState
+import org.yakindu.sct.model.sexec.naming.INamingService
 
 class StateConfVectorIndexCalculator {
+	@Inject extension INamingService
 	protected Map<ExecutionState, Integer> indexMap
+	protected Map<ExecutionState, String> defineMap
 	
 	def Integer getIndex(ExecutionState it) {
 		if(indexMap == null) {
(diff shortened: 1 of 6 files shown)


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StateConfVectorIndexCalculator.xtend
```diff
@@ -0,0 +1,46 @@
+package org.yakindu.sct.generator.c
```
andreasmuelder: Missing Copyright header

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StateConfVectorIndexCalculator.xtend
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.inject.Inject
+import java.util.HashMap
+import java.util.Map
+import org.yakindu.sct.model.sexec.ExecutionState
+import org.yakindu.sct.model.sexec.naming.INamingService
+
+class StateConfVectorIndexCalculator {
```
andreasmuelder: Missing author tag

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StateConfVectorIndexCalculator.xtend
```diff
@@ -0,0 +1,46 @@
...
+	@Inject extension INamingService
+	protected Map<ExecutionState, Integer> indexMap
+	protected Map<ExecutionState, String> defineMap
+	
+	def Integer getIndex(ExecutionState it) {
```
andreasmuelder: I would remove the Map. Its just a call to stateVector.offset, there is no need to cache the result for this. Then the IndexCalculator can be used stateless and as a singleton.

terfloth: I agree!
- The class does not calculate anything as the calculation of the index is already calculated.
- it is effectively just a shortcut access extension that allows to write `it.index` instead of `it.stateVector.offset`.
- `it.stateVector.offset` is more precise about what is really meant.

So - please remove this class

PR #1367, opened 2017-05-10
Check for correct initialization in isActive function (C Generator)

### LINKED ISSUE #1318: isActive() not working on a non initialised machine
Hi,

The isActive() generated method is like this 
```
typedef enum
{
	main_region_Waiting_table_installation,
	FSM_Lifecycle_last_state
} FSM_LifecycleStates;

sc_boolean FSM_Lifecycle::isActive()
{
	return stateConfVector[0] != FSM_Lifecycle_last_state;
}
```

The "[machine]States" enum generated with each machine doesn't have any default value. Hence, the constructor will in most case take the first state as the default value.

In my case  : stateConfVector[0] = 0 = main_region_Waiting_table_installation

A call to the init() method will prevent the problem to happen as it initialisation the state to FSM_Lifecycle_last_state.

If I can understand that it's working if the user following the API contract, I would expect a more robust behavior from Yakindu in case init() is never called, as in complex project it's quite difficult to find such a bug.

I would suggest to add a default invalid value with label 0 to state enums and to use it instead of FSM_Lifecycle_last_state which is not done for that.

```
typedef enum
{
        FSM_Lifecycle_default,
	main_region_Waiting_table_installation,
	FSM_Lifecycle_last_state
} FSM_LifecycleStates;

sc_boolean FSM_Lifecycle::isActive()
{
	return stateConfVector[0] != FSM_Lifecycle_default;
}
```

BeckmaR:
Resolves #1318.

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StatemachineHeader.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StatemachineHeader.xtend
@@ -124,7 +124,7 @@ class StatemachineHeader implements IContentTemplate {
 		/*! Enumeration of all states */ 
 		typedef enum
 		{
-			NOT_INITIALIZED_STATE, /* State with value 0 is the initial value of the StateConfVector /*
+			NOT_INITIALIZED_STATE, /* State with value 0 is the initial value of the StateConfVector */
 			«FOR state : states»
 				«state.shortName»,
 			«ENDFOR»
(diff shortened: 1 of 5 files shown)


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/StatemachineHeader.xtend
```diff
@@ -123,6 +124,7 @@ class StatemachineHeader implements IContentTemplate {
 		/*! Enumeration of all states */ 
 		typedef enum
 		{
+			NOT_INITIALIZED_STATE, /* State with value 0 is the initial value of the StateConfVector */
```
terfloth: - The enum value does not follow the existing naming pattern. Also the 'last state' value uses the stjatemachine name prefix to avoid name clashes.
- the state should be named something like 'XYZ_NO_STATE' 
- Is there any objection to replace the 'lastState' value by the 'no state' (value 0 instead of #states)

PR #1378, opened 2017-05-18
Issue pro 694

andreasmuelder:


plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/transformation/StructureMapping.xtend
```diff
@@ -72,17 +72,18 @@ class StructureMapping {
 	 * Import scope has imports which needs to be resolved to get all imported variable and operation definitions
 	 */
 	def dispatch Scope mapScope(ImportScope scope) {
+		//FIXME!!!
```
BeckmaR: Should be a `#TODO` which is indexed by Eclipse automatically... And maybe some more comment why the code was removed and what needs fixing.

PR #1380, opened 2017-05-18
Introduced named parameters

andreasmuelder:


plugins/org.yakindu.base.expressions/emf-gen/org/yakindu/base/expressions/expressions/util/ArgumentSorter.java
```diff
@@ -0,0 +1,57 @@
...
+					}
+				}
+			}
+		}
+
+		EList<Expression> resultAsList = new BasicEList<Expression>();
+		for (int i = 0; i < result.length; i++) {
+			Expression expression = result[i];
+			if (expression != null)
+				resultAsList.add(expression);
```
tkutz: This part could be problematic. Consider you have only p1 and p3, but not p2 in arguments, then you would return [p1,p3], i.e. p3 value on the position of p2

andreasmuelder: And then the Wrong Number of Arguments Validation Rule should hit

PR #1382, opened 2017-05-19
Provide string based imports

jdicks:
same as #1378 but properly rebased.


plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/transformation/StructureMapping.xtend
```diff
@@ -73,16 +73,17 @@ class StructureMapping {
...
-			val props = resolver.getImportedElementsOfType(imp, Property)
-			for (Declaration decl : props) {
-				_scope.declarations.add(decl.createImportDeclaration)
-			}
-			val operations = resolver.getImportedElementsOfType(imp, Operation) 
-			for (Declaration decl : operations) {
-				_scope.declarations.add(decl.createImportDeclaration)
-			}
-		}
+		//FIXME
```
jdicks: @andreasmuelder how to fix this? 

i can use IPackageImport2URIMapper to get PackageImports - is it needed to get a Package for it?

PR #1391, opened 2017-05-23
isActive returns correct results on inactive statecharts

BeckmaR:
* Changed the order of the state enum, so that the null state is the first one. This makes it the default value and should return correct results on statechart handles that were not initialized. Still, the user should not do that.
* Changed the isActive Function to a for loop, because nobody likes super long lines with twenty **or** operations.


terfloth:
@paphko @DefCon-CC : indeed the fact that 'last_state' is the first one is confusing. Th change was introduced while fixing this issue. From a functional point of view the change makes sens. The 'last_state' was never used as some last index in loops but only in the sense of 'no_state'. And 'no_state' should better have value '0' for a consistent initialization etc. 

So basically it was missed to rename the value to 'no_state'. That should make the intent clear.

PR #1402, opened 2017-05-30
Add check & warning for sgen that does not contain any entries

### LINKED ISSUE #1381: Error / Warning marker on empty *.sgen
Our user Pal Csanyi reported to our User Group that he was not able to generate Code from the statechart ([Forum post](https://groups.google.com/forum/#!topic/yakindu-user/p6PBStIa0WU)).
We finally figured out he did not select a statechart in the sgen creation wizard, so he ended up with an sgen like this:
```
GeneratorModel for yakindu::java {	
}
```
I cannot see any situation where a sgen like this might be useful, so we should consider adding an error / warning marker to sgens that are unable to generate anything, that is:
* completely empty sgens
* sgens with an empty GeneratorModel
Also, depending on the execution internals of the sgen generator, it might be good to print to the console that no action was done.

Without these informations, it can be hard for beginners to track down  why nothing is generated, because they have no idea what an sgen should look like.

BeckmaR:
The following sgen content will now create a warning:

```
GeneratorModel for yakindu::java {
}
```

The text is simply ".sgen file does not contain any entries"

Resolves #1381

### DIFF
--- a/plugins/org.yakindu.sct.generator.genmodel/src/org/yakindu/sct/generator/genmodel/validation/SGenJavaValidator.java
+++ b/plugins/org.yakindu.sct.generator.genmodel/src/org/yakindu/sct/generator/genmodel/validation/SGenJavaValidator.java
@@ -16,6 +16,7 @@ import static com.google.common.collect.Iterables.transform;
 
 import java.util.List;
 
+import org.antlr.grammar.v3.ANTLRv3Parser.finallyClause_return;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.emf.ecore.resource.impl.ResourceSetImpl;
 import org.eclipse.xtext.EcoreUtil2;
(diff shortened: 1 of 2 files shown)


plugins/org.yakindu.sct.generator.genmodel/src/org/yakindu/sct/generator/genmodel/validation/SGenJavaValidator.java
```diff
@@ -148,6 +149,14 @@ public void checkGeneratorExists(GeneratorModel model) {
 					SGenPackage.Literals.GENERATOR_MODEL__GENERATOR_ID);
 		}
 	}
+	
+	@Check
+	public void checkEntriesExist(GeneratorModel model) {
+		if(model.getEntries() == null || model.getEntries().isEmpty()) {
```
andreasmuelder: model.getEntries is never null, EMF one to many references always return empty lists so the == null check is obsolete

PR #1417, opened 2017-06-12
Fixes #1282 Add hint for Ctrl-Space to comment of definition section

### LINKED ISSUE #1282: The hint "Content Assist Available" is too hard to notice
This hint is crucial for a convenient usage of sct but in all of the usability tests not a single user was aware of the lamp-symbol that should inform the user.

To tackle this, there a multiple options:
a) Make the Lamp-Symbol more prominent (i.e. larger in size)
+ Easy to implement
- Might be annoying if you already know about Ctrl+Space

b) Switch to another Symbol which is more noticable but keep the size. For example a bright green exclamation mark
+ Easy to implement
- Might still get overlooked by users

c) Implement it in a way that this hint cannot be overlooked until the user clicks it away
+ This would make sure that the users is aware of this important hint 
- Not entirely show how this can be implemented

svenjawendler:
Please merge this file and close #1282


tkutz:
It is also ctrl space for mac because cmd space is reserved for spotlight search.

Issue #1419, opened 2017-06-12
Filter CycleBased, EventDriven & ExecutionOrder out of c-domain content assist of e.g. variable type


andreasmuelder:
When using the content assist for e.g. variable types the proposals CycleBased, EventDriven & ExecutionOrder are provided, can be inserted and will not be validated even if they do not make sense at this position.

interface:
var pointA : CycleBased

jdicks:
CycleBased & EventDriven aren't visible anymore as expected - ExecutionOrder is still visible

jdicks:
There are 2 lements in Scope called ExecutionOrder:
1. the AnnotationType -> is filtered
2. the EnumerationType -> still visible

andreasmuelder:
This is ok, I can't forbid references to Enumertion Types

PR #1420, opened 2017-06-12
#1419 removes annotations for type proposals.

andreasmuelder:
Scope will provide annotations only if the reference is equals to
annotation_type and not only compatible to annotation_type


andreasmuelder:
@jdicks please remove codacy from your account

jdicks:
@andreasmuelder: removed

PR #1421, opened 2017-06-12
Adds double click drag tracker to notes

### LINKED ISSUE #1392: [GUI] Notes edition is not practical
If a some difficulties to edit notes efficiently in the state machine graphical diagram editor. When I double click on a note the field is not editable, since I would expect it enters edition mode.

At present time I often go quickly in copying current text, deleting the note and creating a new one, which is not the expected workflow.

andreasmuelder:
resolves #1392


andreasmuelder:
@BeckmaR can you check this?

PR #1422, opened 2017-06-12
String-based imports for C domain statecharts

jdicks:


jdicks:
@tkutz : plz have a look

PR #1426, opened 2017-06-13
Null-safe handling of text calculation

tkutz:
Theoretically we could define a property without a name, then we will run into an NPE when, e.g. displaying the element in outline view.

### DIFF
--- a/plugins/org.yakindu.base.types.edit/src/org/yakindu/base/types/provider/PropertyItemProvider.java
+++ b/plugins/org.yakindu.base.types.edit/src/org/yakindu/base/types/provider/PropertyItemProvider.java
@@ -144,10 +144,13 @@ public class PropertyItemProvider
 	 */
 	@Override
 	public String getText(Object object) {
-		Property variable = (Property)object;
-		StringBuilder builder = new StringBuilder(variable.getName());
-		builder.append(" : ");
-		if(variable.getType() != null)
+		Property variable = (Property) object;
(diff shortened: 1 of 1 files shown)


tkutz:
@jdicks please review

plugins/org.yakindu.base.types.edit/src/org/yakindu/base/types/provider/PropertyItemProvider.java
```diff
@@ -144,10 +144,13 @@ public Object getImage(Object object) {
 	 */
 	@Override
 	public String getText(Object object) {
-		Property variable = (Property)object;
-		StringBuilder builder = new StringBuilder(variable.getName());
-		builder.append(" : ");
-		if(variable.getType() != null)
+		Property variable = (Property) object;
+		StringBuilder builder = new StringBuilder();
+		if (variable.getName() != null) {
```
jdicks: is it possible to calculate somekind of sensefgul default in case no name is available?

might 'noName : aType' make sense?

jdicks: or simply ' - : TypeName' ?
or ' : TypeName' ?

tkutz: In our case (SCTUnit) we have properties without names and only want to see the types in the Outline View.

Issue #1393, opened 2017-05-24
[GUI] Composite state inner region scroll


wixiw:
When I click on a composite state inner region, the graphical editor automatically scroll the view (I suppose to try something like a "zoom to fit". It's very annoying as it moves the item on which I have the "mind" focus that I have to search after the scroll. 

I would expect the view to stay on its current position. 

It's an edition performance issue, as it doesn't prevent to reach the user goal.

andreasmuelder:
Not sure if we should change this. Maybe your expectation here is not the same as other users would expect. @svenjawendler what do you think? If we add an option to the diagram preferences, what should be the default?

svenjawendler:
The user only clicks on a region and suddenly the whole diagram moves. For my point of view and also of other users  I talked to (it also came up in the surveys) this is not **conformous to the user expectation** which is a requirement for ergonomic user interfaces (ISO 9041 110). 

My suggestion: 
1. Change the default behavior to not moving the diagram if the user did not explicitly forced it. 
2. If there does not exist a requirement for 'readjusting' the view, I would suggest not to do it.

andreasmuelder:
@svenjawendler thanks, we will implement it as you suggested.

andreasmuelder:
fixed via PR #1428

PR #1428, opened 2017-06-14
disables auto reveal on selection #1393

andreasmuelder:
Change the default behavior to not moving the diagram if the user
selects an editpart


andreasmuelder:
resolves #1428

andreasmuelder:
@svenjawendler can you review this and check if it behaves as expected?

PR #1479, opened 2017-07-05
Reload genmodel

tkutz:
Otherwise, using xcore for meta models that make use of the type meta model shows strange errors with message "A generic super type must refer to a class". The reason for that is not clear, but reloading the genmodel helped.


tkutz:
Also to be noted that re-generating model code with the updated genmodel does not change anything.

Issue #1483, opened 2017-07-06
Missing parameter validation if optional parameters are present


tkutz:
Consider following scenario, an operation with a required and an optional parameter:
```
operation op(pReq : integer, pOpt : integer)
```

Then the following operation call does not give the expected validation error:
```
op(true)
```

PR #1484, opened 2017-07-06
Run type validation for all operation call arguments (#1483)

tkutz:
Since parameters can be optional it is valid to have less arguments than parameters, hence we need to validate all given arguments.


tkutz:
Fixes #1483

tkutz:
As usual, one line of code fixes the issue, but 100 lines are required for the test ;)

PR #1487, opened 2017-07-07
Enable validation also for generator tests

tkutz:
We should discuss if this is a solution. 
Actually it makes sense to skip validation for generator tests, because otherwise we would have also generator errors when we have a regression in validation for example.


andreasmuelder:
better a failing build with an unclear error msg as a non failing build

PR #1490, opened 2017-07-11
Newly generated the CTests and adjusted the GTestHelper

jthoene:
The adjustment was necessary because of the virtual timer service for
the generated sctunit tests


jthoene:
Closed, will not be merged for this Release

PR #1498, opened 2017-07-13
Documentation: references to "import" statement commented out

RainerKlute:


plugins/org.yakindu.sct.doc.user/src/user-guide/sctunit.textile
```diff
@@ -957,6 +976,6 @@ p(#sctunit_fig_test_unit_grammar).
 
 !images/sctunit_grammar_030_test_unit.png(Test unit grammar)!
 
-p=. Test unit grammar
+p=. Test unit grammar %{display: block; font-size: 50%; padding-left: 20%; padding-right: 20%;}*Please note:* The _PackageImport_ non-terminal symbol will be available in future versions of YAKINDU Statechart Tools.%
```
jthoene: Not sure whether it will be available or not in future versions

jthoene: Therefore it would be better if we use an updated graphic without the PackageImport

tkutz: How much effort is it to take it out?

Issue #1438, opened 2017-06-20
Validate Statechart Annotations


BeckmaR:
``` 
@CycleBased(200)
@EventDriven
```
"Works"

terfloth:
Validate exclusive existence of 
* @CycleBased vs EventDriven
* @ChildFirstExecution vs ParentFirstExecution

terfloth:
Check for ChildFirst vs ParentFirst is pending...

PR #1552, opened 2017-07-26
Add validation for co-existing CycleBased & EventDriven

BeckmaR:
See #1438 - but ChildFirst / ParentFirst is still missing, because those don't exist yet on master.

Looks like this - currently no way to underline the annotations was found b/c they don't "exist" in the model.
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: java
+dist: precise
 jdk: 
  - oraclejdk8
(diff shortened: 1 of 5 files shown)


andreasmuelder:
@BeckmaR The build failed due to missing tests:
STextJavaValidatorTest.testAllChecksHaveTests:542 Cannot find test method for checkAnnotations(Statechart)

PR #1553, opened 2017-07-26
Yakindu/sctpro#986: editing domain accessible

jdicks:
* ensures a proper editing domain can be initialized from outside


andreasmuelder:
This is obsolete with the new cache implementation. closing

PR #1582, opened 2017-08-02
Gtest gtesthelper extensions & newly generated C/Cpp Tests

jthoene:
Relevant commits:

f920588

4bc7599


jthoene:
@andreasmuelder : did the requested changes, please review again

BeckmaR:
Could we maybe avoid having PR's with several hundred changed files? Especially when not all of them are generated it is absolutely impossible to review the one or two ones that are changed by hand. Then you should maybe make two PR's or at least reference the commits with the handmade changes in your description.

PR #1584, opened 2017-08-02
Use separate editing domain for import resource cache

andreasmuelder:
Otherwise deadlocks may occure when proxy resolution triggers
DiagramDomainResourceChangeListeners


andreasmuelder:
@jdicks This maybe affect the headless integration

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/ImportedResourceCache.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * 
+ * @author andreas muelder - Initial contribution and API
+ * 
+ */
+@Singleton
+public class ImportedResourceCache {
+
+	private static final String DOMAIN_ID = "org.yakindu.sct.domain.resources";
```
jdicks: should be public because we have to access the editing domain during headless execution

PR #1601, opened 2017-08-07
Extend ID creation to take non-named containers into account (#1500)

### LINKED ISSUE #1500: Extend ID creation to take non-named containers into account
Status Quo:
On type model level, every element has an ID which is basically a full qualified name based on the containment hierarchy. This ID will get the URI fragment of the element which is used to find it in a resource. Using qualified names instead of indices based on location allows to find these elements also when these elements get moved within the same resource.

Shortcomings:
When using our base model as base for more complex languages, there is a need to not only take named containers but also unnamed containers into account for the ID/Fragment computation. Consider a simple if/else statement with a variable x in both branches (if and else) then both x need to have a unique ID, something like "if.x" and "else.x".

tkutz:
fixes #1500

### DIFF
--- a/plugins/org.yakindu.base.types/src/org/yakindu/base/types/TypesUtil.java
+++ b/plugins/org.yakindu.base.types/src/org/yakindu/base/types/TypesUtil.java
@@ -10,7 +10,9 @@
  */
 package org.yakindu.base.types;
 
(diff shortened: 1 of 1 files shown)


tkutz:
@andreasmuelder please review

PR #1632, opened 2017-09-04
Rename parameter in setTimer from timer to timerInterface - fix #1612

### LINKED ISSUE #1612: C++ shadowing errors for member variables
We got following request:

> Problem:
>  
> In member function 'virtual void MyTimedStateMachine::setTimer(TimerInterface*)':
> build/yakindu/Release/board/MyTimedStateMachine.cpp:NNN:NN: error: declaration of 'timer' shadows a member of ' MyTimedStateMachine ' [-Werror=shadow]
>  
> Solution:
>  
> Prefix data members with “m_” to make the data member’s name m_timer.

BeckmaR:
With `-Wshadow` gcc complained about the member `TimerInterface * timer` being shadowed by the setter parameter `TimerInterface * timer`. This was no real issue of course, because the member variable is referenced with `this->timer`, but it was easy to fix with renaming the parameter from `timer` to `timerInterface`.

Fixes #1612

### DIFF
--- a/plugins/org.yakindu.sct.generator.cpp/src/org/yakindu/sct/generator/cpp/StatemachineHeader.xtend
+++ b/plugins/org.yakindu.sct.generator.cpp/src/org/yakindu/sct/generator/cpp/StatemachineHeader.xtend
@@ -303,7 +303,7 @@ class StatemachineHeader extends org.yakindu.sct.generator.c.StatemachineHeader 
 		/*
 		 * Functions inherited from TimedStatemachineInterface
 		 */
-		virtual void setTimer(«timerInterface»* timer);
+		virtual void setTimer(«timerInterface»* timerInterface);
 		
 		virtual «timerInterface»* getTimer();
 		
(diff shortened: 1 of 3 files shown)


terfloth:
Does that solution really solve the problem? What if the user defines a variable called 'timerInterface'?

BeckmaR:
That would not be a problem, because these variables would be members of a nested interface child class. Here, the parameter `TimerInterface * timer` shadows the member `TimerInterface * timer` of the main statechart class.

terfloth:
Perhaps we should apply a consistent C++ code style like that from Google: https://google.github.io/styleguide/cppguide.html#Variable_Names  

Here member variables are named with trailing '_' in this case it would be 'timer_'. Of course there may be others...

BeckmaR:
@terfloth Do you really want to do this in this sprint? I think we'd need more time for that, so maybe we should merge this now and talk about naming changes later?

terfloth:
How many member variables would be affected if we would apply it consistently for all?

Issue #1634, opened 2017-09-05
Effective C++ comliant code - Three warnings occur by using timer


rherrmannr:
Fix compiler warning if you are using -Weffc++ compiler flag:

```
warning: ... has pointer data members [-Weffc++]
warning:   but does not override ... [-Weffc++]
warning:   or 'operator=(const ...)' [-Weffc++]
```

This error occurs, because the timer gets declared as a pointer:
`` TimerInterface* timer; `` 

Therefore, a copy constructor and an assignment operator must be declared.

rherrmannr:
Easiest fix by using Item 6:
Declare copy constructor and assignment operator private. 

see "Effective C++" 
``Item 6: Explicitly disallow the use of compilergenerated
functions you do not want.`` 

This works, but then it's no more possoble to assign or copy statechart objects:
```
Default d;
Default b;
b=d 
```
```
Default d;
Default b(d);
```

rherrmannr:
Using proposals of Item 11
``Item 11: Handle assignment to self in operator=.``

results an implementation like this:
```
Default& Default::operator=(const Default& rhs)
{
	TimerInterface *temp = timer;
	timer = new TimerInterface(*rhs.timer); // this line leads to compile errors
	delete temp;

	return *this;
}
```
Which leads to compile errors. Pure virtual methods of the timer interface must be implemented first.

rherrmannr:
Copy constructor could be implemented like this:

``Default::Default(const Default&rhs):timer(rhs.timer),stateConfVectorPosition(rhs.stateConfVectorPosition),iface(rhs.iface){ }`` 

My suggestion:
Declare assignment operator private and copy constructor public.

BeckmaR:
I don't really like the idea to implement methods that literally no one needs just to conform with some book. Wouldn't it maybe be easier to remove the pointer member?

rherrmannr:
@terfloth what's your opinion?

elgeekomm:
Thanks, this is helpful! Our codebase is Weffc++ compliant and we had to disable this feature for our generated Statecharts, which is something we would like to avoid.

PR #1648, opened 2017-09-13
Issue 1634

rherrmannr:


plugins/org.yakindu.sct.generator.cpp/src/org/yakindu/sct/generator/cpp/StatemachineImplementation.xtend
```diff
@@ -98,26 +101,51 @@ class StatemachineImplementation implements IContentTemplate {
...
+	'''
+		«module»::«module»():
+			«initialisationList»
 		{
 			«constructorBody(it)»
 		}
 	'''
+	}
+	
+	def protected copyConstructerDefinition(ExecutionFlow it) {
```
BeckmaR: This is never called, is it? Please remove dead code

rherrmannr:
fix #1634

rherrmannr:
@terfloth 
Using pointers does not have any effect on the -Weffc++ warnings

This does not result any warnings:
```
#include <stdint.h>

class Pointer{
public:
	class Sub{
	private:
		int16_t *a;
	};
private:
	Sub iface;
};
```

I've also tested it in the product.

Issue #422, opened 2016-02-18
Improve simulation view


SanniSch:
The simulation view is not prominent enough. Therefore there should be some improvements:
- Switch the position of the outline view and the simulation view
- The simulation view must be wider (currently only the a part of the first column is visible) -> the first and the second column should be visible
- The width of the columns should adapt automatically to the content
- The simulation buttons (start simulation, stop, pause,...) should be integrated at the top of the simulation view

terfloth:
Derive tasks

svenjawendler:
@SanniSch  Is this ok for you?
![simulationperspective](https://cloud.githubusercontent.com/assets/22005752/25616374/768ed90e-2f3c-11e7-8d87-113c679227b5.png)

terfloth:
* We should check if the simulation view layout is smart enough to adapt to view space. There should be no minimal size requirement.

SimonEit:
A follow-up suggestion for this issue:

During usability tests users reported that the current way the transitions are listed in the simulation view reminds them more of "hyperlinks" than actual buttons. Light blue underlined text appears as a link to an external source.

A smarter way would be to display the transitions as visual clickable buttons.

robert-rudi:
simulation view has been improved and new features are already merged.

PR #1649, opened 2017-09-13
Added column resizing in simulation view. Both columns (Name, Value) …

robert-rudi:
…are now always visible. The "Value" column has always 30% width within the simulation view. related #422


plugins/org.yakindu.sct.simulation.ui/src/org/yakindu/sct/simulation/ui/view/ExecutionContextViewerFactory.java
```diff
@@ -86,6 +89,60 @@ public void beforeEditorActivated(ColumnViewerEditorActivationEvent event) {
...
+			Point oldSize = tree.getSize();
+			if (oldSize.x > area.width) {
+				setColumnWidths(vBar, width);
+				tree.setSize(area.width, area.height);
+			} else {
+				tree.setSize(area.width, area.height);
+				setColumnWidths(vBar, width);
+			}
+		}
+		protected void setColumnWidths(ScrollBar vBar, int width) {
```
BeckmaR: I like this, except the point that the value and name columns will always be scaled to ⅓ / ⅔ when the size of the whole value view thing is changed, no matter how the user set it up before. I think it is a good initial value but overriding the user's choices on each resizing is kind of drastic - especially because this also appplies when the **window** is resized. Maybe it would be better to keep the user's proportions. You never know what kind of long enum literal someone wants to display in the value column.

robert-rudi: I've built in your suggestion in the following commit. The user's proportions should now be taken into account, even if the window itself is resized.

andreasmuelder:
created as new PR

PR #1655, opened 2017-09-19
Changed save prompt for "Run as..." option for a statechart project …

robert-rudi:
…which is dirty. Saving of project can now be set to be automatically saved.


plugins/org.yakindu.sct.simulation.core/src/org/yakindu/sct/simulation/core/launch/AbstractSCTLaunchConfigurationDelegate.java
```diff
@@ -1,90 +1,101 @@
...
+		return false;
+	}
+
+	@Override
+	protected IProject[] getBuildOrder(ILaunchConfiguration configuration, String mode) throws CoreException {
+		return getProjectsForProblemSearch(configuration, mode);
+	}
+	
+	@Override
+	protected boolean saveBeforeLaunch(ILaunchConfiguration configuration, String mode, IProgressMonitor monitor)
```
andreasmuelder: Can be removed

PR #1677, opened 2017-09-22
Documentation: section "Generating SCTUnit source code" marked as being a PRO feature

RainerKlute:


andreasmuelder:
It is not a pro feature, it is also available in the Standard Edition.

PR #1678, opened 2017-09-25
C/C++ generator using modulo operators on sc_real (double)

### LINKED ISSUE #1654: C/C++ generator using modulo operators on sc_real (double)
include math.h and use fmod(double,double) instead of % operation

rherrmannr:
fix #1654


rherrmannr:
fixed with duplicated pr

Issue #1333, opened 2017-05-01
OCB interfaces should be checked for null in enter() method


wixiw:
I'm using an embedded environment without any RTTI available, so NULL pointer are not catched. It's quite painfull to have the board stucked in an HW error each time I forgot to map a new OCB interface.

It would be nice to check the interface mapping in the init() or enter() method.

terfloth:
OK - checking the operation callbacks makes much sense...

andreasmuelder:
the same applies to  the Java Code Generator. The operation callbacks should be checked in the enter method. Init could break existing client code.

public void enter() {
		if (!initialized) {
			throw new IllegalStateException(
					"The state machine needs to be initialized first by calling the init() function.");
		}
		
		if(sCInterface.operationCallback == null) {
			throw new IllegalStateException("Operation callback for interface 'scInterface' must be set");
		}
		
		enterSequence_main_region_default();
	}

rherrmannr:
We could implement it like this:
```
void Default::enter()
{
#ifdef CHECK_UNIMPLEMENTED_VIRTUAL_METHODS
	// Check if methods are implemented
	// if not
	throw std::runtime_error( "Operation callback for interface XY must be set.");
#endif
	/* Default enter sequence for statechart default */
	enseq_main_region_default();
}
```

and define CHECK_UNIMPLEMENTED_VIRTUAL_METHODS in the StatemachineInterface.h file. 
It can be deleted, if some IDE's are not supporting #include <stdexcept>

What's your opinion @terfloth ?

wixiw:
This wouldn't answer the need expressed at the issue opening. No RTTI being present you cannot use exceptions. If they where there, I would have had an exception on the NULL pointer. 

I may agree that generating a clean runtime error is better than a NULL pointer exception, but it's still using exceptions.

What would fit is a return code to the enter method. It should not break existing code as the function is currently returning a void which is by definition not tested by user code.

terfloth:
We have to add the checks to ::init() as operation callbacks may be used during variable initialization. Example:

```javascript
operation getSomeValue(): integer
var x : integer = getSomeValue()
```

rherrmannr:
Implemented in Java within the init() function.

rherrmannr:
As an example I've implemented defines and changed the return type of the init function into the cpp generator.

The first four bits indicates, that an OCB is unimplemented. The four right handed bits indicates, which type of OCB is unimplemented. This can be discussed:
[...]

PR #1702, opened 2017-10-10
OCB interfaces should be checked for null in enter() method 

rherrmannr:
issue https://github.com/Yakindu/statecharts/issues/1333 only for java


terfloth:
This issue is not correctly implemented!! See issue #1333

PR #1718, opened 2017-10-18
fix EMF Compare preference labels

Bananeweizen:
The custom processors for EMF Compare all need a label, otherwise the
class names are displayed in the preference pages.

Feel free to change the wording.


andreasmuelder:
Thx!
I just changed 'state chart' to 'statechart' according to our naming convention.

PR #1723, opened 2017-10-23
Refactor generators

BeckmaR:
Introduces a new language model to design the statechart class in one step and create the textual representation in another. This allows a much easier inheritance, because adding, editing and removing function objects from a statechart class object is much more elegant than changing already generated string representations. All new classes inherit from a new class named CodePart which in turn implements CharSequence. __Thus, all new classes can be used in Xtend templates and intermixed with regular strings and stuff.__

Another improvement: The class only has to be built once, with all its member variables and functions, and can then be declared and implemented without further work. This prevents errors like declaring a function as `int` but implementing as `void` and similar from happening.

Logic is split between a base collection of stuff, mainly a bunch of interfaces, found in `org.yakindu.sct.generator.core.language` and implementations for the language - that is currently C and C++ - in a respective `.language` sub-package. When more generators are moved to the new model, it would probably possible to move even more stuff to the base implementation and make the language specifics slimmer.


BeckmaR:
Closing, needs to be split into smaller parts and integrated over time.

PR #1729, opened 2017-10-25
[SCTUnit] Assertions like "assert integer == integer" are not allowed 

rherrmannr:
For example: assert boolean == boolean is not allowed

fix https://github.com/Yakindu/sctpro/issues/1223


andreasmuelder:
should be added to the SCTUnit Valdiator since this is not possible in SText / Expressions

PR #1731, opened 2017-10-26
Compile error in CPP code for string assignments within conditional expressions

### LINKED ISSUE #1594: Compile error in CPP code for string assignments within conditional expressions
To reproduce, take digital_watch example from example repository and generate CPP code from it.

You will find some errors, one of which is this one:

bildschirmfoto 2017-08-04 um 12 23 10

Normal assignments to string variables work fine. 
It seems like CPP behaves different regarding type conversion when the string is inside another expression, here a conditional expression.

rherrmannr:
fix #1594


rherrmannr:
Added test for conditional expressions with strings 
conditionalexpressions

PR #1746, opened 2017-11-16
Issue 1684

### LINKED ISSUE #1684: Folding/Unfolding of transitions is not working in combination with rename refactoring
The folding/unfolding mechanism is not working after a name of a variable/etc. is refactored. The folding/unfolding then does nothing, except enabling/disabling those operations in the context menu UI only.
(2) Also folding/unfolding become disabled, after the value of the refactored variable is manually changed.  

Steps to reproduce:
        
- Create a statechart with two states (e.g. LightOff, LightOn)
- Create a transition from state LightOff to state LightOn with the expression:
        `A.keypress / light = true`
- Replace the content in the interface definition section with the following:
```
internal:
var light : boolean
interface A:
in event keypress

```
- Try fold/unfold the outgoing/incoming transition for each of the states.
- The folding/unfolding should work correctly.
- Refactor the name of the variable 'light' to any arbitrary name (e.g. 'lamp').
- After the renaming, the fold/undold mechanism doesn't work anymore.
(2)
- Change name value of the refactored name (i.e. if lamp was true, change to false, or otherwise)
- Right click one of the states -> the corresponding Fold/Unfold must not be disabled.

robert-rudi:
Fixes #1684

### DIFF
--- a/plugins/org.yakindu.sct.model.sgraph/src/org/yakindu/sct/model/sgraph/resource/AbstractSCTResource.java
+++ b/plugins/org.yakindu.sct.model.sgraph/src/org/yakindu/sct/model/sgraph/resource/AbstractSCTResource.java
@@ -343,6 +343,16 @@ public abstract class AbstractSCTResource extends XMIResourceImpl {
 			}
 		}
 	}
+	
+	public void parseSpecificationElements() {
+		TreeIterator<EObject> iter = getAllContents();
+		while (iter.hasNext()) {
+			EObject currentObject = iter.next();
+			if (currentObject instanceof SpecificationElement) {
+				parseSpecificationElement((SpecificationElement) currentObject);
+			}
+		}
+	}
 
 	protected void linkSpecificationElement(SpecificationElement element) {
 		isLinking = true;
@@ -368,8 +378,10 @@ public abstract class AbstractSCTResource extends XMIResourceImpl {
 				serializeStatechart((Statechart) element);
 			}
 		} catch (XtextSerializationException ex) {
+			ex.printStackTrace();
 			// Leave the old specification
 		} catch (IConcreteSyntaxValidator.InvalidConcreteSyntaxException ex) {
+			ex.printStackTrace();
 			// Leave the old specification
 		} finally {
 			isSerializing = false;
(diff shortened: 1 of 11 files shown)


plugins/org.yakindu.sct.model.sgraph/src/org/yakindu/sct/model/sgraph/resource/AbstractSCTResource.java
```diff
@@ -343,6 +343,16 @@ public void linkSpecificationElements() {
 			}
 		}
 	}
+	
+	public void parseSpecificationElements() {
+		TreeIterator<EObject> iter = getAllContents();
+		while (iter.hasNext()) {
+			EObject currentObject = iter.next();
+			if (currentObject instanceof SpecificationElement) {
+				parseSpecificationElement((SpecificationElement) currentObject);
```
andreasmuelder: can we add an iter.prune() here?

robert-rudi: If a SpecificationElement is a CompositeState, the State can contain further States and Transitions. If a refactoring made changes to those elements then they would't be parsed with iter.prune(), won't they?

plugins/org.yakindu.sct.refactoring/src/org/yakindu/sct/refactoring/refactor/AbstractRefactoring.java
```diff
@@ -129,13 +121,32 @@ protected void executeCommand(AbstractTransactionalCommand refactoringCommand) {
...
 		return true;
 	}
+	
+	/**
+	 * Combines the internal command execution call with the parsing and linking of the specification elements.
+	 */
+	protected void doInternalExecute() {
+		internalExecute();
+		parseAndLinkSpecificationElements();
+	}
```
tkutz: Two things:
1) Usually "doX..." Methods are sub methods of "X.." methods, means that "X" is calling "doX" internally, here it is the other way around. If we change that, we need to adjust all subclasses to implement the "doInternalExecute" method instead of "internalExecute".

2) Now that refactoring execution is combined with parse and link functionality we can remove these calls from the tests!

plugins/org.yakindu.sct.refactoring/src/org/yakindu/sct/refactoring/refactor/AbstractRefactoring.java
```diff
@@ -129,13 +121,32 @@ protected void executeCommand(AbstractTransactionalCommand refactoringCommand) {
...
 
 	/**
 	 * Is called within {@link IRefactoring#execute()} method.
 	 */
 	protected abstract void internalExecute();
 
 	/**
+	 * Parsed and links the specification elements of {@link AbstractRefactoring#getResource()}
```
tkutz: Parse**s**...

plugins/org.yakindu.sct.model.sgraph/src/org/yakindu/sct/model/sgraph/resource/AbstractSCTResource.java
andreasmuelder: yes you are right

PR #1755, opened 2017-11-22
OCB interfaces should be checked for null in enter() method

### LINKED ISSUE #1333: OCB interfaces should be checked for null in enter() method

rherrmannr:
fix #1333


rherrmannr:
will be fixed with issue_1333_dupl

PR #1784, opened 2017-12-11
Remove bundle dependency to org.apache.log4J, fixes #1768

### LINKED ISSUE #1768: Hard plugin-dependency to org.apache.log4j instead of import-package
To be able to change the log4j implementation (e.g. by the implementation log4j.over.slf4j) it is necessary to not use Require-Bundle but instead use Import-Package.

But in the MANIFEST.MF of org.yakindu.sct.model.stext this is done.

Can you please change this?

BeckmaR:


andreasmuelder:
@BeckmaR we should a require bundle dependency then

andreasmuelder:
okay, it is already there, merging

PR #1833, opened 2018-01-02
reset zoom with CTRL '0'

### LINKED ISSUE #1528: [Ctrl+0] should reset the zoom factor
While [Ctrl++] and [Ctrl+-] zoom in and out as expected, the user is missing [Ctrl+0] to reset the zoom factor, which he is used to in web browsers and other software.

andreasmuelder:
resolves #1528


tkutz:
Added [Ctrl] + [Numpad 0] also.

PR #1838, opened 2018-01-03
init enumeration types with the first enum literal

andreasmuelder:


andreasmuelder:
#resolves https://github.com/Yakindu/sctpro/issues/1281

PR #1844, opened 2018-01-05
Issue 1562: Parent-first and child-first and their relationship to local reactions explained

RainerKlute:


plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -1311,6 +1269,250 @@ bc. myBool = active(StateA)
...
+h4. Child-first/parent-first and local reactions
+
+The following examples are somewhat more complex. The statechart used in the examples defines the integer variables _m_, _n_, _o_, _p_, _q_, and _r_. Initially, they are all set to 0. Local reactions in states *A* and *B* set these variables to non-zero values on certain conditions:
+* when entering a state,
+* when leaving a state, and
+* when receiving event _e_.
+
+This is useful to understand which actions are performed (or not performed) and in which order.
+
+The local reactions in composite state *A* are as follows:
```
terfloth: entry and exit actions play a specific role within state transitions and are no local reactions even though they are defined in the same location. Please also be aware that you can combine the triggers like ' e, entry, exit / doSomething' . In this case a local reaction as well as entry and exit actions are defined. (It is not neccessary to explain this here....)
  
You could write something like 'The reactions defined in composite state "A" ...."

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -1311,6 +1269,250 @@ bc. myBool = active(StateA)
...
+
+bc. entry / m = 1
+e / n = 1
+exit / o = 1; r = 1
+
+The local reactions in substate *B* are as follows:
```
terfloth: See above ...

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: Ah, I wasnt aware that entry/exit reactions don't count as "local" reactions! On the other hand, does such a differentiation really matter to the user (or to anyone else)? Wouldn't it be easier for everyone to define all reactions that are "inside the box" as being "local reactions"?

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: I removed all occurrences of "local".

PR #1857, opened 2018-01-12
Issue pro 1293

BeckmaR:
Introduces a new duplicate name validator which is able to catch:

```
interface:
var a: integer

interface a:
```

as a duplicate NamedElement, as well as producing the error message "Duplicate State" instead of "Duplicate NamedElement" - by better comparing the EClasses and super classes of the clashing objects.


plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextNamesAreUniqueValidationHelper.java
```diff
@@ -0,0 +1,176 @@
...
+		return node;
+	}
+
+	protected EClass checkForCommonSuperClass(IEObjectDescription one, IEObjectDescription two,
+			ValidationMessageAcceptor acceptor) {
+		
+		Node<EClass> superClassesOne = getSuperClassTree(one.getEClass());
+		Node<EClass> superClassesTwo = getSuperClassTree(two.getEClass());
+		
+		List<EClass> flatOne = superClassesOne.flatten();
```
andreasmuelder: If you use only the flat representation of the super type hierachy there is no need to build a graph data structure (Node<T>).

PR #1890, opened 2018-02-05
place regions for extracted sub diagrams beside each other

andreasmuelder:
also fixed a bug when state text expression editpart was selected when
executing proposal provider.


andreasmuelder:
@svenjawendler can you have a look at this?

PR #1901, opened 2018-02-08
WIP Definitionsection with annotationborder and rename refactoring support

robert-rudi:
- Replaced the simple texteditor for the definition section with an embedded Xtext editor which has an annotation border for error/warning marker.
- Enabled RenameRefactoring in new definition section editor
- Fixed NPE in STextNameAreUniqueValidationHelper which was caused after specification was saved (due to fakeresource usage in StatechartDiagramEditor).


robert-rudi:
Fixes #1922

PR #1902, opened 2018-02-08
Fix #1899

### LINKED ISSUE #1899: Regression: Spec Sec warning underlining missing in some conditions
Apparently, the yellow squiggly underlining is missing in the specification section under certain conditions, for example for "unused declaration in the internal interface". This feature worked in earlier builds.

BeckmaR:
This fixes #1899 where some issues failed to be displayed in the definition section text. The reason seems to be that checks that are not `CheckType.FAST` were filtered from the definition section.


BeckmaR:
Doesn't what I thought it would, closing this.

PR #1913, opened 2018-02-15
Fix EventDriven Simulation behavior

### LINKED ISSUE #1880: Simulation skips handling out-events in event-driven mode
See the animation below. 
In cycle-based mode, state B is activated. In event-driven mode, it is not.

This is probably a bug in simulator, where out-events are discarded too early.

eventdrivenregions

BeckmaR:
Prior to this change, out events are not raised when evaluated in the same
RunCycle, because they are put in the internalEventQueue and then ignored.

Furthermore, out events should not start a RunCycle, because that is not useful
and OutEvents are cleared on the beginning of a RunCycle.

Fixes #1880


BeckmaR:
@terfloth tests implemented

PR #1929, opened 2018-02-26
Asserting outgoing events documented

RainerKlute:
Assertion of outgoing events documented, fixing YSCTpro-1243.

Additionally:
* Railroad charts updated
* Minor improvements


plugins/org.yakindu.sct.doc.user/src/user-guide/sctunit.textile
```diff
@@ -299,9 +299,9 @@ bc.
 
 ==</div>==
 
-p. The _header_ of an operation consists of an optional annotation, typically <code>&#x40;Test</code>, followed by the keyword @operation@, followed by the name of the operation, followed by a parenthesized list of parameters, optionally followed by a colon and a return type.
+p. The _header_ of an operation consists of an optional annotation, typically <code>&#x40;Test</code>, followed by the keyword @operation@, followed by the name of the operation, followed by a parenthesized list of parameters, optionally followed by a colon and a return type. Specifying no return type is equivalent to explicitly specifying @void@ as return type.
```
andreasmuelder: I think this is not the case, when no type is specificed it will be inferred by its return statements, like

operation doit() {
return 5
}

var x : integer = doIt() is valid

RainerKlute: Thanks, fixed.

PR #1949, opened 2018-03-19
Extending the Naming class to also support naming for sgraph elements

jthoene:
This is necessary for the generation of SCTUnit Code which isn't based
on the ExecutionFlow model but based on the Statechart (sgraph) model.


jthoene:
@andreasmuelder : please add a review this is kinda urgent for the progress of the SCTUnit C Generator.

PR #1960, opened 2018-03-20
added docu for c++ fileextensions

rherrmannr:


andreasmuelder:
better add this to a more general topic than C++ classes

PR #1963, opened 2018-03-20
Validation for PostFix

### LINKED ISSUE #1957: Post-incr on boolean is not validated, but simulation crashes
title. You can do:

```
var counter: boolean
```

and then `counter++`, which does not give any errors, but the simulation crashes.

andreasmuelder:
resolves #1957


plugins/org.yakindu.base.expressions/src/org/yakindu/base/expressions/validation/ExpressionsJavaValidator.java
```diff
@@ -123,6 +127,14 @@ public void accept(ValidationIssue issue) {
 		}
 	}
 
+	@Check
+	public void checkPostFixOperatorOnlyOnVariables(PostFixUnaryExpression expression) {
+		if (!(expression.getOperand() instanceof ElementReferenceExpression)
+				&& !(expression.getOperand() instanceof FeatureCall)) {
```
tkutz: One of both is redundant

tkutz: Sorry, didnt see the !

PR #1999, opened 2018-04-02
Statechart smells detection (both sgraph and stext)

herrendorf:
PR for statechart smells.

A small testsuite is also included


andreasmuelder:
thanks a lot for this contribution I will have a detailed look at this next week.

andreasmuelder:
SgraphJavaValidationTest fails because of code smells ;-)
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.yakindu.sct.model.sgraph.test.SGraphJavaValidationTest
log4j:WARN No appenders could be found for logger (org.eclipse.xtext.ui.internal.Activator).
log4j:WARN Please initialize the log4j system properly.
Tests run: 44, Failures: 5, Errors: 1, Skipped: 0, Time elapsed: 4.058 sec <<< FAILURE!

andreasmuelder:
I think that this rule is too restrictive. 
From my point of view it is valid to have a state machine with just one State and a local reaction.
Initial --> StateA [clicked / counter++]
However, this could make sense for sink states without local reaction. In both cases an existing exit action could be validated, too. @terfloth what is your opinion on that?

BeckmaR:
We could add a statechart annotation that activates these tests, I think that would be nice to have.

BeckmaR:
@herrendorf it would be nice if you could remove german comments / translate them and remove commented code :)

herrendorf:
I fixed the comments and removed the unused code.

Do you implement / provide that annotation or should I try doing it myself?

herrendorf:
I did further improvements and added an informal description of the smells as JavaDoc.
I disabled `checkNoNonFinalStateIsSink` by commenting `@Check` out. So the CI Build should work again ;)

herrendorf:
Oh, the CI build is still broken. I'll check that on the weekend.

herrendorf:
Since I just used some sample Statecharts to test my smells, they aren't covered by unit tests. Thus 
`org.yakindu.sct.model.sgraph.test.SGraphJavaValidationTest.testAllChecksHaveTests(SGraphJavaValidationTest.java:841)` fails.
If I disable this test on my local machine, everything is fine. But of course I don't want to commit this change ;)

I also broke lots of stext tests. Maybe we can talk about that first on friday, before I start digging through the tests.

andreasmuelder:
Closing this, as discussed, all smell detections are contributed as separate tickets

PR #2005, opened 2018-04-05
remove QID import style

andreasmuelder:


andreasmuelder:
closes #https://github.com/Yakindu/sctpro/issues/1429

PR #2006, opened 2018-04-05
xtext runner package migration

andreasmuelder:


andreasmuelder:
not supported for Mars platform

PR #2032, opened 2018-04-13
Issue 2020: EventDriven Java Generator

### LINKED ISSUE #2020: Remove unused functions from generated event driven Java code.
Same as #2016

BeckmaR:
Only tiny fixes, because the java generator already barely contained any useless code when no local events were present.

Fixes #2020


BeckmaR:
Currently does not build because SCT-CI is outdated, and cannot be rebuilt right now because the Neon target can't be fetched.

PR #2038, opened 2018-04-16
used local ScopedElement instead of persistent statechart

### LINKED ISSUE #2037: [Scoping] Statechart Definition Section ignores local scope
The local dirty state of the definition section is not used, instead the persistent state is always used.

How to reproduce:

open definition section editor, type:
internal:
var test : boolean
var test2 : boolean = <CTRL-SPACE>

the variable 'test' can not be referenced. The editor  has to be saved until it is referenceable

andreasmuelder:
resolves #2037


robert-rudi:
For the attached examples the wrong scopes are used.
javakeywords_error
exitstate_error
ckeywords_error

So I think we must use the scopes of the statechart itself relative from the scoped element we are looking at.

Issue #2045, opened 2018-04-21
always, oncycle, every and after may be used as names for 'in events'


herrendorf:
The internal event names `always`, `after`, `oncycle` and `every` may be used as names for `in events`:

```
@CycleBased(200)

interface:
// Define events and
// and variables here. 
//Use CTRL + Space for content assist.

in event always 
```

This causes an ambiguity, if the event is used on  transition.

herrendorf:
`after` and `every` already break the parser and cause a `STextParser$SyntaxEception`. So an error is shown, but the message is rather confusing:
```
No viable alternative at input 'after'
Qualified name cannot be null```

andreasmuelder:
same for exit and entry triggers

andreasmuelder:
fixed by https://github.com/Yakindu/statecharts/pull/2057

PR #2046, opened 2018-04-21
Issue 2045

herrendorf:
Fix for issue #2045: Add error marker to broken event definition.


andreasmuelder:
changed base to issue_2045 to check this on grammar side

Issue #2047, opened 2018-04-21
Smell: All outgoing transitions of a state or a choice have the same effect


herrendorf:
If all outgoing transitions of a state or a choice have identical effects (including guards), the effect could be defined in the state using `onexit` or choice's incoming transitions.

Might not be useful, if the choice has more than one incoming transitions. Opinions?

Issue #2048, opened 2018-04-21
Smell: All incoming transitions of a state have identical effects


herrendorf:
analogue of #2047 for incoming transitions:
 
If all incoming transitions of a state or a synchronization have identical effects (including guards), the effect could be defined in the state using `entry` or synchronization's outgoing transitions.

Again, this might not be useful, if a synchronization has lots of outgoing transitions.

andreasmuelder:
What if we exclude the synchronizations and only add this check for States?

BeckmaR:
Or it's only active for syncs when they have only one outgoing transition.

herrendorf:
I prefer @BeckmaR's idea and will update my PR.

PR #2049, opened 2018-04-21
Issue 2047

herrendorf:
Check to detect smell #2047


herrendorf:
As in #2048 if restricted the check for the 'choice' node to the case, when it has one incoming transition.

herrendorf:
I could provide a sct file with a statechart, that contains that smell. Or should I try to implement the testcase by constructing the statechat programmatically?

andreasmuelder:

PR #2050, opened 2018-04-21
Issue 2048

herrendorf:
Implemented a check to detect smell #2048 . Unittest is only a stub.


herrendorf:
I could provide an sct file with a statechart, that contains that smell. Would this be useful?

andreasmuelder:

Issue #2053, opened 2018-04-22
Smell: Using the same trigger multiple times on a transition is redundant


herrendorf:
A transition with reaction
`always, oncycle, myEvent, myEvent [someGuardExpression] / someActions`
contains redundant information and might also be expressed as
`always, myEvent [someGuardExpression] / someActions`

Note: Multiple `after` triggers (e.g. `after 2s, after 10s`) are only considered, if there's no guard. If there's a guard, it's expression might be false after 2s, but can be true after 10s => This is a valid (but not obvious) use case.

herrendorf:
That's not possible. The guard 'guards' _all_ triggers at the same time. It's not possible to specifiy multiple guards on one transition:
```
myEvent [false], myEvent [true]
```
raises an syntax error (see also [Yakindu docs](https://www.itemis.com/en/yakindu/state-machine/documentation/user-guide/sclang_statechart_language_reference#sclang_trigger_specifications) syntax diagram)

PR #2054, opened 2018-04-22
Issue 2053

herrendorf:
Added a check for #2053 

There are some problems, where I need further information:
 * What's the meaning of `every` on a transition? It doesn't make sense for me, but it's possible. (Even if a comment in my code claims the opposite)
 * The unit test is `@Ignore`d, because the condition at the beginning of `checkDuplicateTriggers()` evaluates to false in the way, the expression is created during the unit test. Is there an (easy) way to fix this?
* The smell is only reasonable, if `@CycleBased()` is set. How can I check that via code?

### DIFF
--- a/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
+++ b/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
@@ -17,12 +17,14 @@ import static org.yakindu.sct.model.stext.lib.StatechartAnnotations.EVENT_DRIVEN
 import static org.yakindu.sct.model.stext.lib.StatechartAnnotations.PARENT_FIRST_ANNOTATION;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


andreasmuelder:
@herrendorf regarding your questions:

> What's  the meaning of every on a transition?
Every only makes sense with a guard. (Could ne another smell, every without guard should be replaced with an after.) Such a trigger makes sense: 'Every 10 s [x == 10]' every 10 seconds the guard is evaluated.

> The smell is only reasonable, if @CycleBased() is set. How can I check that via code?
 org.yakindu.sct.model.stext.lib.StatechartAnnotations#isCycleBased

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
```diff
@@ -957,7 +958,7 @@ protected boolean isResponsible(Map<Object, Object> context, EObject eObject) {
 		return true;
 	}
 
-	protected Statechart getStatechart(EObject context) {
+	protected static Statechart getStatechart(EObject context) {
```
herrendorf: I made this static, because it can be static and I make all my helper methods static, if it's possible.

herrendorf:
I broke some Unit tests: `org.yakindu.sct.model.stext.test.TestModelsContainErrorsTest.testTestModelContainsErrors()`
I've no idea, what I did ...

andreasmuelder:

PR #2064, opened 2018-05-04
Modularize generators

BeckmaR:
Split the C generator into multiple parts, removed inheritance to `StatemachineHeader` and `StatemachineSource`. Contains also changes from #2059 which should be merged first.


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/HeaderContentFragmentProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ * @author rbeckmann
+ *
+ */
+public class HeaderContentFragmentProvider implements Provider<StatemachineHeader> {
+	@Inject
+	Injector injector;
+
+	@Override
+	public StatemachineHeader get() {
+		StatemachineHeader source = new StatemachineHeader();
```
andreasmuelder: why not injector.getInstance(StatemachineHeader.class) here?
if 'new' is used here binding overriding will not work!

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/SourceContentFragmentProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Provider;
+
+/**
+ * @author rbeckmann
+ *
+ */
+public class SourceContentFragmentProvider implements Provider<StatemachineSource> {
```
andreasmuelder: This class is similar to HeaderContentFragmentProvider you should extract an AbstractContentFragmentProvider

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/HeaderContentFragmentProvider.java
BeckmaR: I can tell you why - it leads to StackOverflowErrors to get an instance via injector in the provider that is bound in the module. ;) This provider should always provide StatemachineHeader, nothing else, and you would rebind it to another provider.

BeckmaR:
If this gets merged, the version must be bumped to 3.5.0 due to incompatibilities.

BeckmaR:
Closing this as this was basically done with #2175

PR #2076, opened 2018-05-17
Change initial statechart decl to use eventdriven and childfirst

### LINKED ISSUE #2073: New Statechart Models should be created with @EventDriven and @ChildFirst per default
We should add a short text or a Text Note to the with a hint to this change and a link to the docu.

Maybe we could also extend the Wizard with a selection for the different annotations

BeckmaR:
Resolve #2073


plugins/org.yakindu.sct.ui.editor/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,4 +1,14 @@
 eclipse.preferences.version=1
+org.eclipse.jdt.core.codeComplete.argumentPrefixes=
```
andreasmuelder: Please remove JDT preferences from the project since we use the settings defined in oomph setup file

BeckmaR: These are only to enable Eclipse to understand the comments `formatter:on` and `formatter:off`, because without it the statechart decl template will be made unreadable by the auto formatter...

andreasmuelder: you should add this settings to oomph setup then, formatter preferences should not be project specific

BeckmaR:
@andreasmuelder merged master, removed project local settings

PR #2078, opened 2018-05-18
SCTPro issue 1468

jthoene:
This fixes https://github.com/Yakindu/sctpro/issues/1468

Returning the result after the evaluation and not the result before the
evaluation.


andreasmuelder:
Please add a test to STextInterpreterTest

PR #2083, opened 2018-05-23
WIP Change SCT design

BeckmaR:
This PR modernizes the look of SCT a bit:

* Color gradients in states and regions are removed
* Shadows of states are removed
* Regions are now painted plain white
* The canvas is grey (before, these two colors were swapped)
* The background of transition expressions is white and opaque, so that it covers the transition when put on top of it
* The default color of numbers in expressions when using syntax coloring was darkened to improve the contrast with the new state color (see `42`)

Preview:
image


tkutz:
A side by side comparison would be nice. Also once you play with colors & shadows, showing some alternatives makes a decision easier. Also when talking about colors here, we need to see how the simulation (and breakpoint) highlighting color looks.

I for myself would like to see this new colour-sheme with some shadows, e.g. inline shadows for regions, so it looks like a region is carved in a state, and normal shadows for states, so the hierarchy of states gets more graspable.

BeckmaR:
@tkutz inset shadows look weird:

image

terfloth:
I agree we should have a side by side comparison.

I'm so sad that you want to remove the gradient as it was so much fun to implement it ...  ;-)

two suggestions: 
*  Instead of using a grey i would propose to use  slightly darker blue. This will provide the same degree of contrast but makes the design even 'flatter'. If the user has set a custom  color we have to calculate that color.  
* I would not use the shadowed inset. Instead we could try a simple slim 'lowered bevel border' as an alternative design.
* If we remove gradients i at least want to see some animations ;-)

BeckmaR:
New theme:

image

old theme:

image

andreasmuelder:
Don't forget the black theme : https://groups.google.com/forum/#!topic/yakindu-user/kDl3JJzPX6s

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/figures/TransitionFigure.java
```diff
@@ -6,27 +6,29 @@
...
  * @author muelder
- * 
+ *
  */
 public class TransitionFigure extends PolylineConnectionEx {
 
 	private final IMapMode mapMode;
 
-	protected static final int TOLERANCE = 4;
+	protected static final int TOLERANCE = 3;
```
andreasmuelder: why that?

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/themes/AbstractTheme.java
```diff
@@ -0,0 +1,95 @@
+package org.yakindu.sct.ui.editor.editor.themes;
```
andreasmuelder: EPL

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/themes/ColorHelper.java
```diff
@@ -0,0 +1,171 @@
+package org.yakindu.sct.ui.editor.editor.themes;
+
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.RGB;
+
+public class ColorHelper {
+	public static double getShadePercentage(Color color) {
+		HSL hsl = HSL.fromRGB(color.getRGB());
```
andreasmuelder: Is this HSL color space really required? Something that can not be done with RGB + libs?

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/figures/TransitionFigure.java
BeckmaR: I think that was a mistake, I will remove it.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/editor/themes/ColorHelper.java
BeckmaR: Yeah, HSL allows to transform colors programmatically which is not really possible in RGB, and there doesn't seem to be a good library offering HSL conversion.

BeckmaR:
There is one major issue: we can't differentiate between state colors the user chose and the default colors. This means there is no way to recolor existing statecharts when the user selects a different theme without discarding the user's color choices, e.g. for error states or whatever. Selecting a theme without the possibility to recolor old statecharts is useless, so we have to think more about this.

In theory, the information if it's a default color or not is hidden in the xmi. In the first case, the color is only attached to the `StateFigureCompartment`, while a user chosen color is also attached to the `State` element.

tuttoaldoc:
I agree that is very important to have the possibility to apply the chosen theme also to previous statecharts. We would like to have a black theme for development and a white one for printing. When are you planning to work on this subject? It's a pitty this recoloring issue.
Best Regards
Michele

BeckmaR:
Due to dependencies of the statechart editor, there is currently no good way to offer a consistent experience or a black theme. closing

PR #2085, opened 2018-05-28
fixed datatype conversion for serialized simulation context

andreasmuelder:


plugins/org.yakindu.sct.model.sruntime/src/org/yakindu/sct/model/sruntime/SRuntimeFactoryImplOverride.java
```diff
@@ -0,0 +1,47 @@
...
+ * 
+ * @author andreas muelder - Initial contribution and API
+ * 
+ */
+public class SRuntimeFactoryImplOverride extends SRuntimeFactoryImpl implements SRuntimeFactory {
+
+	@Override
+	public Object createJavaObjectFromString(EDataType eDataType, String initialValue) {
+		int indexOf = initialValue.indexOf(":");
+		String className = initialValue.substring(0, indexOf);
```
tkutz: Are we sure that, `indexOf > 0` always holds? If not, an exception is thrown here. I would be more defensive.

PR #2105, opened 2018-06-26
Partly reverted the StextExtension because of side effects

jthoene:
The Injection of the methods corrupted the SCTUnit Scope because of
inheritance dependencies as well as led to exceptions when using the
editor ui


tkutz:
Closed because this is reworked in https://github.com/Yakindu/statecharts/pull/2107

PR #2107, opened 2018-06-26
Created Interface for Extensions

jthoene:
Also injected this interface in every class which had a getStatechart
method which now uses the method getStatechart provided by the
interface. Bounded the STextExtension for the SText context.

### DIFF
--- a/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/extensions/ISCTExtensions.java
+++ b/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/extensions/ISCTExtensions.java
@@ -1,3 +1,13 @@
+/**
+ * Copyright (c) 2018 committers of YAKINDU and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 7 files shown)


plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/extensions/ISCTExtensions.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface ISCTExtensions {
+	public SpecificationElement getContextElement(EObject context);
+	
+	public ScopedElement getScopedElement(EObject context);
+	
+	public Statechart getStatechart(EObject context);
+	
+	public Statechart getStatechart(final Resource context);
+}
```
tkutz: License header and author tags

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/STextGlobalScopeProvider.java
```diff
@@ -77,6 +76,8 @@
 	private IPackageImport2URIMapper mapper;
 	@Inject
 	private ImportedResourceCache resourceDescriptionCache;
+	@Inject
+	private ISCTExtensions utils;
```
tkutz: could also inject as extension, then you can write "resource.getStatechart"

PR #2146, opened 2018-07-20
YSCT 2145: Generating documentation with YAKINDU Docs now

RainerKlute:


BeckmaR:
I guess this will never work on Travis because the linked jar repositories are not accessible from the outside...

PR #2175, opened 2018-08-09
Refactoring of C-Code-Generator to support Mynewt-Generator extensions.

terfloth:
Makes the C-Code Generator more extendable.


tkutz:
I ran the C generators for following example projects to compare the generated code:
* coffee_machine_pro
* coffee_machine_pro_cpp
* traffic_light_cpp

The differences found are whitespaces and order of includes/define/functions. These are fine so far. Only ugly thing I encountered was wrong indentation of switch-case blocks (left is the mynewt branch, which has too much indentation for "break" and also for closing brackets in the end):

bildschirmfoto 2018-08-09 um 11 18 51

tkutz:
Okay, fixed the switch-case indentation "bug".

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/EventNaming.xtend
```diff
@@ -19,6 +20,7 @@ import org.yakindu.sct.model.sexec.naming.INamingService
 /**
  * @author René Beckmann
  */
+@Singleton // Guice
```
tkutz: Why is this required?

PR #2208, opened 2018-09-03
"Readonly" is deprecated but helpful

### LINKED ISSUE #2192: "Readonly" is deprecated but helpful
In the latest Version's of Yakindu SCT the variable "prefix" readonly is marked as deprecated. But this is helpful if you would like to prevent that a variable can only be read from outside of the state machine and never been written.

rherrmannr:
fix #2192


rherrmannr:
@RainerKlute please check the changed user docu part (statechart_language.textile)

PR #2211, opened 2018-09-04
update shallow history docs

tangele:


plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -445,31 +445,30 @@ p=. Orthogonal states
...
 
-!images/docu_shallowHistory02.jpg(Shallow history [2])!
 
-p=. Shallow history &#91;2]
+!images/docu_shallow_history_pause.png(Shallow history [4])!
 
-!images/docu_shallowHistory03.jpg(Shallow history [3])!
+When the _continue_ event is triggered, the most recent active state inside of the composite state _answeringQuestions_ is activated again. The respondent doesn't need to answer the previous questions again, he can simply resume.
 
-p=. Shallow history &#91;3]
```
tkutz: @RainerKlute what was this for, what do we miss if we remove these lines?

tkutz:
LGTM

@RainerKlute can you have a look at this?

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -445,31 +445,30 @@ p=. Orthogonal states
 
 h3(#sclang_shallow_history_states). Shallow history states
 
-A *shallow history state* is a pseudo state that is placed inside the region of a composite state. It is used to remember the last active state inside a composite state. This makes it possible to jump back to the remembered state instead of starting at the inner initial state again.
+A *shallow history state* is a pseudo state that is placed inside the region of a composite.
```
BeckmaR: I do prefer the old version. The new one is confusing.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: These lines are image captions.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
RainerKlute: I restored the removed phrase as it tells the reader what a shallow history state is good for. It is a necessary step on the ladder of understanding and thus should not be left out.

plugins/org.yakindu.sct.doc.user/src/user-guide/statechart_language.textile
```diff
@@ -445,31 +445,30 @@ p=. Orthogonal states
...
 ==<div class="example">==
 
-###. FIXME: Re-create this example with nice names and even nicer drawings!
-
 The following example, showing the answering of a questionaire, explains this:
 
-!images/docu_shallowHistory01.jpg(Shallow history [1])!
+!images/docu_shallow_history_init.png(Shallow history [1])!
+!images/docu_shallow_history_nextQuestion1.png(Shallow history [2])!
+!images/docu_shallow_history_nextQuestion2.png(Shallow history [3])!
```
RainerKlute: Filenames are case-sensitive! Please always use lower-case, including the filename extension!

RainerKlute: I also optimized the image file sizes.

RainerKlute: The images need visual improvents in order to look nice:
* The *pause* transition arrow should end in the middle of the **init** state, not at its low-left corner.
* The **q** states should be of the same size.
* The **q** states should be properly aligned, e.g., vertically.
* The **init** state should have the same size as the **q** states.

plugins/org.yakindu.sct.model.sruntime/META-INF/MANIFEST.MF
```diff
@@ -1,21 +0,0 @@
-Manifest-Version: 1.0
```
tangele: Is the deletion of this file intentional?

PR #2215, opened 2018-09-04
Cleanup NamesAreUniqueValidationHelper & ignore statechart name

### LINKED ISSUE #2188: Duplicate NamedElement in Statechart
I'm getting nowadays an error in Yakindu SCT 3.3.2 that duplicate I have a duplicate named element in the Statechart.

The problem comes from the fact that the name of the State-Machine is the same as one interface in the Statechart.

As this was never a problem with older versions of Yakindu SCT it could be that this an error

BeckmaR:
Fixes #2188

### DIFF
--- a/plugins/org.yakindu.sct.commons/src/org/yakindu/sct/commons/EMFHelper.java
+++ b/plugins/org.yakindu.sct.commons/src/org/yakindu/sct/commons/EMFHelper.java
@@ -1,8 +1,23 @@
+/**
+ * Copyright (c) 2018 committers of YAKINDU and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ * Contributors:
+ * 	committers of YAKINDU - initial API and implementation
+ *
+ */
 package org.yakindu.sct.commons;
+
 
 import java.util.ArrayList;
 import java.util.List;
 
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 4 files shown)


plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextNamesAreUniqueValidationHelper.java
```diff
@@ -73,66 +71,58 @@ public void checkUniqueNames(Iterable<IEObjectDescription> descriptions, CancelI
...
-		createDuplicateNameWarning(description, description.getEClass(), acceptor);
-		createDuplicateNameWarning(lowerCasePut, description.getEClass(), acceptor);
 	}
 
-	protected void duplicateFQN(IEObjectDescription description, IEObjectDescription old, EClass common,
-			ValidationMessageAcceptor acceptor) {
-		createDuplicateNameError(description, common, acceptor);
-		createDuplicateNameError(old, common, acceptor);
+	protected boolean shouldValidateEClass(EClass eC) {
+		if (eC.isSuperTypeOf(SGraphPackage.Literals.STATECHART)) {
```
andreasmuelder: return !eC.isSuperTypeOf(SGraphPackage.Literals.STATECHART)

plugins/org.yakindu.sct.commons/src/org/yakindu/sct/commons/EMFHelper.java
```diff
@@ -0,0 +1,35 @@
+package org.yakindu.sct.commons;
```
andreasmuelder: can't you just use 
EClass clazz;
clazz.isSuperTypeOf(someOtherClass)
And delete EMFHelper

BeckmaR: Nope - I want to know if `B extends A` and `C extends A` are related, but they are not super types of each other.

PR #2233, opened 2018-09-10
Generator executor return IStatus for further processing of errors

tkutz:


plugins/org.yakindu.sct.generator.core/src/org/yakindu/sct/generator/core/execution/AbstractGeneratorEntryExecutor.java
```diff
@@ -51,16 +55,23 @@
...
 		logStart(entry);
 		if (valid(entry)) {
 			try {
 				execute(factory.create(entry), entry);
 			} catch (Exception ex) {
 				logger.logError(ex);
+				status = new Status(IStatus.ERROR, PLUGIN_ID, ex.getMessage());
 			}
+		} else {
+			status = new Status(IStatus.ERROR, PLUGIN_ID,
```
andreasmuelder: is Status.ERROR correct here since there is no error occured during generation. Maybe Status.Warning or Status.Cancel is betteR?

tkutz: At this point the model validation has found an error in the referenced statechart. These errors are also already printed out on console log. So it is a validation error here.

PR #2250, opened 2018-09-14
testcases: add expectation for additional warning message

tangele:


BeckmaR:
Why is this not to be merged since two weeks? What's the current state?

PR #2265, opened 2018-09-19
added validation for choices and regular states for triggers

### LINKED ISSUE #505: Implement priority adaption for default trigger on regular states
From forum post by *johnm*:

> In the doc I read this:
> 
> default, else
> 
> The default trigger is equivalent to the else trigger. It is intended to be used for the outgoing transitions of choice pseudo states to make sure that there is always an outgoing transition that can be taken. The default trigger can only be be used in transitions and implies the lowest evaluation priority for that transition. 
> 
> 
> But when I defined two transitions from a simple state (not a choice pseudostate) and assumed that the one with the else trigger would only happen if the other one didn't, I was caught out. I had to set the correct sequence manually on the state's Properties pane.

Applying the existing priority adaption to regular states could make sense.

rherrmannr:
fix #505 

using always/oncycle on choices should not be used

using default/else/always/oncycle should have the lowest priority for
regular states

### DIFF
--- a/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
+++ b/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
@@ -68,6 +68,7 @@ import org.yakindu.sct.model.sgraph.Entry;
 import org.yakindu.sct.model.sgraph.Exit;
 import org.yakindu.sct.model.sgraph.ReactionProperty;
 import org.yakindu.sct.model.sgraph.Region;
+import org.yakindu.sct.model.sgraph.RegularState;
 import org.yakindu.sct.model.sgraph.SGraphPackage;
 import org.yakindu.sct.model.sgraph.Scope;
 import org.yakindu.sct.model.sgraph.ScopedElement;
@@ -83,6 +84,7 @@ import org.yakindu.sct.model.stext.extensions.STextExtensions;
 import org.yakindu.sct.model.stext.scoping.IPackageImport2URIMapper;
 import org.yakindu.sct.model.stext.scoping.IPackageImport2URIMapper.PackageImport;
 import org.yakindu.sct.model.stext.services.STextGrammarAccess;
+import org.yakindu.sct.model.stext.stext.AlwaysEvent;
 import org.yakindu.sct.model.stext.stext.ArgumentedAnnotation;
 import org.yakindu.sct.model.stext.stext.DefaultTrigger;
 import org.yakindu.sct.model.stext.stext.EntryEvent;
@@ -124,6 +126,8 @@ import com.google.inject.name.Named;
 		STextNamesAreUniqueValidator.class })
 public class STextJavaValidator extends AbstractSTextJavaValidator implements STextValidationMessages {
 
+	private static final String KEYWORD_ONCYCLE = "oncycle";
+	private static final String KEYWORD_ALWAYS = "always";
 	@Inject
 	private ITypeSystemInferrer typeInferrer;
 	@Inject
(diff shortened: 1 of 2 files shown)


rherrmannr:
retest this please

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextValidationMessages.java
```diff
@@ -61,4 +61,6 @@
 	public static final String CONTRADICTORY_ANNOTATIONS = "Some annotations (%s) have contradictory effects.";
 	public static final String BAD_EVENT_NAMES = "'%s' is already used as name of an internal event.";
 	public static final String SYNC_OUTGOING_TRIGGER = "Triggers and guards on a synchronization's outgoing transitions will be ignored.";
+	public static final String ALWAYS_TRUE_TRANSITION_USED = "Transition with '%s' should have the lowest priority, otherwise transitions with lower priority will never be taken.";
```
terfloth: This message might be confusing. The highest priority has the lowest number... Additionally state how to fix the problem.

So we should use a alternative message like: "Transition with '%s' should be evaluate as the last transition of the state as following transitions won't be evaluated. Change the transition order (in the Properties view)."

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
```diff
@@ -199,7 +203,63 @@ public void transitionsWithNoTrigger(Transition trans) {
 			warning(ISSUE_TRANSITION_WITHOUT_TRIGGER, trans, null, -1);
 		}
 	}
+	
+	@Check(CheckType.FAST)
+	public void checkAlwaysTransitionHasLowestPriority(RegularState state) {
```
terfloth: Also add warnings for all transitions that have lower priority. Use message like:

' Dead transition. This transition can not be taken due to previous transition with '%s' trigger.

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextValidationMessages.java
```diff
@@ -61,4 +61,6 @@
...
+	public static final String USE_DEFAULT_TRIGGER_IN_CHOICES = "Transition with '%s' effect blocking other transitions. Use 'default' or 'else' instead.";
```
terfloth: The problem is nearly the same as for the other check. So use a comparable message like: 

"Transition with '%s' should be evaluate as the last transition of the state as following transitions won't be evaluated. Use 'default' or 'else' instead or change the transition order (in the Properties view)."

Be aware that the use of 'always' and 'default' are conflicting if both are defined. In this case the 'default' transition will never be taken. This is an additional case that has to be considered.

terfloth: This are two different properties that we have to check. 
1. There should be no dead transitions.
2. It must always be possible to exit a choice.

Taking the existing semantics we can identify the following cases:
a. Using always not with lowest priority violates 1). 
b. Using always with lowest priotity and there is no additional default fulfills 1) and 2)
c. Using always and default together will always violate 2) at least for the default transition.
d. Using two or more default transition also leads to dead transitions (violates 2))

I derive the following conclusions from this: 

i)  a) is exactly the same problem as for regular states. An same problem should be reported with the same message/reason. An additional possible fix in the context of choices is to convert always to default or to fix the order as for regular states.

ii)  Converting always to default is invalid if there is already a default transition - always and default are always conflicting.

iii) if b) is fulfilled than the warning that you should use a choice should not appear. But currently it will appear.

iv) There must be an additional check for multiple defaults.

terfloth: @rherrmannr: please fix iii)

PR #2282, opened 2018-09-24
Make pinned DefSec update name on statechart name change

### LINKED ISSUE #2213: Definition Section and Properties View can get out of sync
image

This can happen when the statechart name is changed in the definition section, sometimes [CTRL] + [Z] is needed, etc.

Also this does persist even when the statechart is saved, so the actual current name is not 100% clear until the statechart is closed and reopened again.

BeckmaR:
Fix #2213

### DIFF
--- a/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
+++ b/plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
@@ -150,7 +150,7 @@ public class StatechartDefinitionSection extends Composite implements IPersistab
 		this.embeddedEditor = createSpecificationEditor();
 		registerResizeListener();
 		GridLayoutFactory.fillDefaults().numColumns(2).spacing(0, 0).applyTo(this);
-		initSynchronizer(DiagramPartitioningUtil.getDiagramContainerHierachy(getDiagram()));
+		initSynchronizer();
 	}
 
 	protected Label createSwitchControl() {
@@ -181,11 +181,13 @@ public class StatechartDefinitionSection extends Composite implements IPersistab
 		return labelComposite;
 	}
 
-	protected void initSynchronizer(List<Diagram> diagramContainerHierachy) {
+	protected void initSynchronizer() {
 		synchronizer = new DefinitionSectionSynchronizer();
-		for (Diagram diagram : diagramContainerHierachy) {
-			diagram.getElement().eAdapters().add(synchronizer);
-		}
+		getContextObject().eAdapters().add(synchronizer);
+	}
+
+	protected void removeSynchronizer() {
+		getContextObject().eAdapters().remove(synchronizer);
 	}
 
 	protected void createNameLabel(Composite labelComposite) {
@@ -446,6 +448,7 @@ public class StatechartDefinitionSection extends Composite implements IPersistab
 	public void dispose() {
 		saveState(memento);
 		disposeEmbeddedEditor();
+		removeSynchronizer();
 		super.dispose();
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
```diff
@@ -766,4 +783,30 @@ public void dispose() {
...
+	 */
+	protected class DefinitionSectionSynchronizer extends AdapterImpl {
+		@Override
+		public void notifyChanged(Notification notification) {
+			if (Notification.SET == notification.getEventType()) {
+				Object feature = notification.getFeature();
+				if (BasePackage.Literals.NAMED_ELEMENT__NAME.equals(feature)) {
+					suppressModifyEvent = true;
+					nameLabel.setText(notification.getNewStringValue());
+					suppressModifyEvent = false;
```
tkutz: Wouldn't it be enough to just check if new name and old name are different, and only set if both are not equal, to avoid endless notification cycles?

BeckmaR: Well, this would still trigger the whole notification cycle twice, which might have some implications now or in the future.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
```diff
@@ -174,15 +181,25 @@ protected Composite createDefinitionSectionComponents() {
 		return labelComposite;
 	}
 
+	protected void initSynchronizer(List<Diagram> diagramContainerHierachy) {
+		synchronizer = new DefinitionSectionSynchronizer();
+		for (Diagram diagram : diagramContainerHierachy) {
+			diagram.getElement().eAdapters().add(synchronizer);
```
andreasmuelder: the statechart is the same for multiple subdiagrams. Only register the adapter once!! on the statechart

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
```diff
@@ -174,15 +181,25 @@ protected Composite createDefinitionSectionComponents() {
...
-		Text nameLabel = new Text(labelComposite, SWT.SINGLE | SWT.NORMAL);
+		nameLabel = new Text(labelComposite, SWT.SINGLE | SWT.NORMAL);
 		GridDataFactory.fillDefaults().indent(5, 1).grab(true, false).align(SWT.FILL, SWT.CENTER).applyTo(nameLabel);
 		nameLabel.setText(getStatechartName());
 		nameLabel.setEditable(isStatechart());
 		nameLabel.setBackground(ColorConstants.white);
 		nameModificationListener = new ModifyListener() {
 			@Override
 			public void modifyText(ModifyEvent e) {
+				if (suppressModifyEvent) {
```
andreasmuelder: hacky af

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
```diff
@@ -174,15 +181,25 @@ protected Composite createDefinitionSectionComponents() {
 		return labelComposite;
 	}
 
+	protected void initSynchronizer(List<Diagram> diagramContainerHierachy) {
+		synchronizer = new DefinitionSectionSynchronizer();
+		for (Diagram diagram : diagramContainerHierachy) {
+			diagram.getElement().eAdapters().add(synchronizer);
+		}
```
andreasmuelder: where is the adapter removed? remove in dispose method.

plugins/org.yakindu.sct.ui.editor/src/org/yakindu/sct/ui/editor/definitionsection/StatechartDefinitionSection.java
```diff
@@ -174,19 +180,35 @@ protected Composite createDefinitionSectionComponents() {
...
+		if (name.isPresent()) {
+			nameLabel.setText(name.get());
+		}
 		nameLabel.setEditable(isStatechart());
 		nameLabel.setBackground(ColorConstants.white);
 		nameModificationListener = new ModifyListener() {
 			@Override
 			public void modifyText(ModifyEvent e) {
-				if (getContextObject() instanceof Statechart) {
+				Optional<Statechart> sct = getStatechart();
```
andreasmuelder: why not use getStatechartName here?

BeckmaR:
Because we need the statechart - or getContextObject - anyhow

René Beckmann

PR #2305, opened 2018-09-28
In gen model, show path to statechart in content assist and tooltip

### LINKED ISSUE #2304: In generator model, show path to referenced statechart in content assist
This helps to understand which statechart is actually referenced in case there are multiple equally named statecharts in the project.

tkutz:
Fixes #2304


andreasmuelder:
As discussed with @tkutz We should add a validation rule to SGen when multiple elements of the same content type exist on the path

PR #2307, opened 2018-10-01
Fix import-string Content Assist

BeckmaR:
The function which created the import proposals just called the wrong method
to create the actual proposal which was an illegal shortcut.

`doCreateProposal` just returns a new Proposal obj, whereas
`createCompletionProposal` calls `isValidProposal` beforehand.


plugins/org.yakindu.sct.model.stext.ui/src/org/yakindu/sct/model/stext/ui/contentassist/STextProposalProvider.java
```diff
@@ -272,15 +273,22 @@ public void complete_BOOL(EObject model, RuleCall ruleCall, ContentAssistContext
...
+		PrefixMatcher newMatcher = new PrefixMatcher() {
+			@Override
+			public boolean isCandidateMatchingPrefix(String name, String prefix) {
+				String strippedName = name;
+				if (name.startsWith("^") && !prefix.startsWith("^")) {
+					strippedName = name.substring(1);
+				}
+				return context.getMatcher().isCandidateMatchingPrefix(strippedName, prefix);
+			}
+		};
```
tkutz: remove if this is not relevant

BeckmaR:
@tkutz done

BeckmaR:
@tkutz Test is done, Github doesn't update label for some reason.

PR #2331, opened 2018-10-10
Use multibindings for operation mockups

tkutz:
This allows to combine different mockups. The first one that is responsible for an operation will be executed.


plugins/org.yakindu.base.expressions.interpreter/src/org/yakindu/base/expressions/interpreter/DefaultExpressionInterpreter.xtend
```diff
@@ -265,11 +269,11 @@ class DefaultExpressionInterpreter extends AbstractExpressionInterpreter impleme
 	}
 
 	def executeElementReferenceExpression(ElementReferenceExpression expression) {
-		var parameter = expression.expressions.map(it|execute)
+		val parameter = expression.expressions.map(it|execute)
 		if (expression.operationCall || expression.reference instanceof Operation) {
-			if (operationDelegate !== null &&
-				operationDelegate.canExecute(expression.reference as Operation, parameter.toArray)) {
-				return (expression.reference as Operation).execute(parameter.toArray)
+			val operationDelegate = operationDelegates.findFirst[canExecute(null, expression.reference as Operation, parameter.toArray)]
```
andreasmuelder: may be null when optional injection = true

PR #2336, opened 2018-10-15
return nullscope if sctunit var or const reference itself

rherrmannr:
fix https://github.com/Yakindu/sctpro/issues/1682


plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/STextScopeProvider.java
```diff
@@ -120,20 +122,24 @@ public IScope scope_ElementReferenceExpression_reference(final EObject context,
...
 
 		Predicate<IEObjectDescription> predicate = calculateFilterPredicate(context, reference);
 
 		Expression owner = context.getOwner();
 		EObject element = null;
 		if (owner instanceof ElementReferenceExpression) {
 			element = ((ElementReferenceExpression) owner).getReference();
+			EObject eContainer = context.eContainer();
+			if(eContainer instanceof VariableDefinition && element instanceof VariableDefinition) {
+				return scope;
```
tkutz: And what happens in a valid case? Do we also return a null scope here?

andreasmuelder:
will not be merged as discussed

PR #2361, opened 2018-10-18
Terminate simulation on project deletion

rherrmannr:
fix https://github.com/Yakindu/sctpro/issues/1662


plugins/org.yakindu.sct.simulation.ui/src/org/yakindu/sct/simulation/ui/model/presenter/SCTSourceDisplayDispatcher.java
```diff
@@ -57,8 +58,20 @@ public void displaySource(Object element, IWorkbenchPage page, boolean forceSour
...
 					(ISimulationEngine) newTarget.getAdapter(ISimulationEngine.class));
 		}
-		activeSourceDisplay.displaySource(newTarget, page, forceSourceLookup);
-		activeDebugTarget = newTarget;
+		
+		// terminate simulation if project has been deleted
+		try {
+			activeSourceDisplay.displaySource(newTarget, page, forceSourceLookup);
+			activeDebugTarget = newTarget;
+		} catch (RuntimeException e) {
```
andreasmuelder: We should not catch RuntimeExceptions, we should check the root cause for the exception instead.

PR #2377, opened 2018-10-23
Pro issue 1460

BeckmaR:
This pull request allows to check imported resources for errors and display them on the import statement.

A good way to sync errors from imported resources with the definition section is missing; the only possible way seems to be to revalidate the definition section when new issues are found on external file changes. This leads to the validation being done twice.


andreasmuelder:
Do not merge for 3.4.2

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
```diff
@@ -899,34 +905,51 @@ public void checkImportExists(ImportScope scope) {
 				error(String.format(IMPORT_NOT_RESOLVED_MSG, packageImport), scope,
 						StextPackage.Literals.IMPORT_SCOPE__IMPORTS, imports.indexOf(packageImport));
 			}
+			else {
+				IResourceDescription iResourceDescription = resourceDescriptionCache.get(pkImport.get().getUri());
+				if (iResourceDescription instanceof DefaultResourceDescription) {
+					Resource resource = ((DefaultResourceDescription) iResourceDescription).getResource();
+					if (!(resource.getWarnings().isEmpty() && resource.getErrors().isEmpty())) {
```
tkutz: Does it mean, I will now get an error in the statechart although the imported resource has just a warning?
I would prefer to not be too strict here. Actually I think an info message would be enough here.

BeckmaR:
@tkutz changes implemented

andreasmuelder:
closing, this will not be part of the 3.x maintainance

PR #2382, opened 2018-10-24
Issue 2380: ModelSequencer

BeckmaR:


BeckmaR:
Tests added with a264c8f actually fail on the current master, so the fixes introduced in this branch also fix the evaluation of transitions.

PR #2384, opened 2018-10-25
improve validation error messages; improve parser failure message "Missing EOF at"

### LINKED ISSUE #2203: check error/warning validation messages for expressiveness and quickfixes
* in a usability test, we noticed that an user (who received the "missing trigger" warning) didn't knew where he was expected to add the missing trigger.
* in the case of the error message ”local declarations are not allowed in interface scope”, the same user was wondered how to fix that. He had defined an event without using the `in` or `out` keyword.

tangele:
fixes #2203


tangele:
not sure if https://github.com/Yakindu/statecharts/pull/2384/commits/57e3ebc1b94f95452f5c34990c72960ea7a97166 is really an improvement. getLinkText() can be null in the Linker implementation,  but i fail to reproduce an example where its null.

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextJavaValidator.java
```diff
@@ -453,11 +453,13 @@ public void checkValueOfNoEvent(EventValueReferenceExpression exp) {
...
 		if (element != null && (!(element instanceof Event))) {
-			String elementName = "";
+			String elementName;
 			if (element instanceof NamedElement) {
 				elementName = "'" + ((NamedElement) element).getName() + "' ";
+			} else {
+				elementName = "Could not find event declaration.";
 			}
-			error(elementName + "is no event.", StextPackage.Literals.EVENT_VALUE_REFERENCE_EXPRESSION__VALUE, 0,
+			error(elementName, StextPackage.Literals.EVENT_VALUE_REFERENCE_EXPRESSION__VALUE, 0,
```
tkutz: In case of NamedElement, error message is only elementName now. Add "is no event" for example.

PR #2406, opened 2018-11-08
Only one 'add statechart' dialog

rherrmannr:
determine domain on projects nature

fix #https://github.com/Yakindu/sctpro/issues/1675


plugins/org.yakindu.sct.ui.editor/META-INF/MANIFEST.MF
```diff
@@ -11,6 +11,7 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.ui.views,
  org.eclipse.ui.navigator,
  org.eclipse.ui.navigator.resources,
+ org.eclipse.cdt.core,
```
andreasmuelder: we can't add a dependency to cdt.core from ui.editor plugin

rherrmannr:
agree

rherrmannr:
Moving to C Domain is not that useful.

Imagine different domains - Java or C/C++.

Other options:
1.We could add the a default option by using a dialog: ``You've used XY Domain. Should we remember your decision?``

2. Have a look into the .project file and search for domain strings

3. Always use default domain...

PR #2408, opened 2018-11-09
NPE in DefaultResourceChangeToIssueProcessor

Bananeweizen:
During SWTBot tests we have seen this exception. It looks like a race
condition in DefaultValidationIssueStore, where that store is being
disconnected (line 160) while another ResourceChangeEvent is still being
processed.

Since I assume that this race condition does not occur under normal
manual usage (but only in fast UI testing) it seems reasonable to just
ignore that (last) resource change event using an additional guard.

java.lang.NullPointerException: null
	at org.yakindu.sct.ui.editor.validation.DefaultResourceChangeToIssueProcessor.getDeltaForFile(DefaultResourceChangeToIssueProcessor.java:65)
~[na:na]
	at org.yakindu.sct.ui.editor.validation.DefaultResourceChangeToIssueProcessor.process(DefaultResourceChangeToIssueProcessor.java:50)
~[na:na]
	at org.yakindu.sct.ui.editor.validation.DefaultValidationIssueStore.resourceChanged(DefaultValidationIssueStore.java:250)
~[na:na]
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:299)
~[na:na]


andreasmuelder:
I have never seen this exception so your assumption seems to be right.  Of course we can merge this for better testability

PR #2423, opened 2018-11-13
Add image to clarify difference between transition and state breakpoints

tkutz:
Fixes #1674


tkutz:
@RainerKlute better now?

RainerKlute:
Much better indeed, @tkutz!

PR #2440, opened 2018-11-21
Remove stdlib.h and string.h from C & C++

### LINKED ISSUE #2388: Only generate includes when used
for C and C++, the headers 'stdlib.h' and 'string.h' are generated as includes into the C Module:
```
#include <stdlib.h>
#include <string.h>
```

We should only generate those includes if they are required

BeckmaR:
Close #2388


terfloth:
@rherrmannr : There is StringExpressions.[sct|sctunit] that covers all string compare flavors.

PR #2445, opened 2018-11-21
Add -I option to makefile compile targets

BeckmaR:
`-I` tells gcc where to look for header files, and was forgotten in the previous
version.

Fix Yakindu/sctpro#1824


BeckmaR:
The code is the original one just modified for StringBuilder. It would be a good test when someone could build a SCT_BRANCH build on Jenkins from this, this should just restore the old behavior as without Makefile.

BeckmaR:
The branch builds on Jenkins, so this can be merged.

PR #2448, opened 2018-11-22
[Java Generator] Add a line wrap between package and import block

tkutz:


BeckmaR:
The method `createImports` is unused now, I just forgot to remove it. Sorry :(

PR #2464, opened 2018-11-28
Extracted ImportUriProvider for subclassing in specific domains

tkutz:
This extracts providing of import URI to an own class to allow for easier subclassing.

### DIFF
--- a/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/ImportUriProvider.java
+++ b/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/ImportUriProvider.java
@@ -1,0 +1,114 @@
+/**
+ * Copyright (c) 2018 committers of YAKINDU and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     committers of YAKINDU - initial API and implementation
+ */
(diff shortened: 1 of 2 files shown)


plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/scoping/ImportUriProvider.java
```diff
@@ -0,0 +1,114 @@
...
+		Optional<PackageImport> pkgImport = mapper.findPackageImport(resource, packageImport);
+		if (pkgImport.isPresent() && pkgImport.get().getUri() != null
+				&& getConverter().exists(pkgImport.get().getUri().trimQuery(), null)) {
+			acceptor.accept(pkgImport.get().getUri().toString());
+		}
+	}
+	
+	protected boolean isValidUri(Resource context, URI uri) {
+		boolean validURI = EcoreUtil2.isValidUri(context, uri);
+		if (!validURI) {
```
BeckmaR: An URI can be invalid but still exist?

andreasmuelder: EcoreUtil2.isValidURI uses the uriConverter from the XtextResourceSet that contains the fake xtext resource. For statechart as a type, we derive a synthetic sct_types resource that only exists in memory so it will return false here. Then we do the check with the URI converter from the transactional editing domain where we contribute a URIConverter for sct_types that can handle synthetic resources properly. A comment would have been a good idea. :)

BeckmaR: magic

PR #2474, opened 2018-12-03
Remove superfluous const in C++ generator

### LINKED ISSUE #2468: const value in SCT description language results in compiler warning
The following SCT expression:
```
const constant_value : integer = 42
```
results in a compiler warning because a const expression is generated for a native return type:
```
./src-gen/src.h:449:43: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
     const sc_integer get_constant_value() const;
```

rherrmannr:
fix #2468

Avoid warnings if -WExtra is enabled.


BeckmaR:
150 files changed :scream_cat:

rherrmannr:
Yes, since we have removed the default comment in SCTUnit and I needed to regenerate the tests. But there is a new Whitespace in the generated .cc-files. I will have a look at this.

rherrmannr:
I will regenerate the tests

PR #2475, opened 2018-12-03
Re-pipe `getAnnotations` to `annotationInfo`

BeckmaR:
Re-pipe `getAnnotations` to `annotationInfo`

Recently (in 714ada2fa46197eb11de39e0f519a8136e5f4965) the grammar was changed and the annotations of `Declaration`s were moved into the field `annotationInfo`. This PR pipes these annotations into `getAnnotationOfType`, because the own field `annotations` never contains anything.

Also, the JDK level of Expressions was set to 8.0.


andreasmuelder:
@BeckmaR can you resolve the conflicts please?

PR #2485, opened 2018-12-04
Use StatechartSpecification for validation

### LINKED ISSUE #2484: checkAnnotations should act on StatechartSpecification, not Statechart
Due to some change, the underlining for error markers does not appear when an error is added to an annotation in the statechart. It works only when the StatechartSpecification is validated and the annotations are taken from there.

BeckmaR:
Fix #2484: For some reason, the error markers on the annotation appear only when they are taken from the `StatechartSpecification`, not the statechart itself.


andreasmuelder:
@BeckmaR can you resolve the conflicts please?

BeckmaR:
@andreasmuelder done

PR #2535, opened 2018-12-13
Cast fmod to reciever to avoid warnings

rherrmannr:
Needed for pro: https://github.com/Yakindu/sctpro/issues/1868

added cast to fmod calls

moved modOnReal check to ExpressionsChecker

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/ExpressionsChecker.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/ExpressionsChecker.xtend
@@ -11,14 +11,18 @@
 package org.yakindu.sct.generator.c.extensions
 
 import com.google.inject.Inject
+import org.eclipse.emf.ecore.EObject
 import org.yakindu.base.expressions.expressions.AssignmentExpression
 import org.yakindu.base.expressions.expressions.AssignmentOperator
+import org.yakindu.base.expressions.expressions.ElementReferenceExpression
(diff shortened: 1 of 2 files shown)


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/ExpressionsChecker.xtend
```diff
@@ -53,4 +60,32 @@ class ExpressionsChecker {
...
+	
+	def boolean haveCommonTypeReal(Expression expression) {
+		if(isSame(getCommonType((infer(expression).getType), getType(ITypeSystem.INTEGER)),
+			getType(ITypeSystem.INTEGER))) return false
+		return true
+	}
+	
+		
+	def CharSequence castToReciever(Expression expression) {
+		if (expression instanceof ElementReferenceExpression) {
```
tkutz: FeatureCall case is missing, so better to use dispatch method here

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/ExpressionsChecker.xtend
```diff
@@ -53,4 +60,32 @@ class ExpressionsChecker {
...
+			val ref = expression.reference
+			if (ref instanceof VariableDefinition) {
+				val originalType = ref.typeSpecifier.type
```
tkutz: Use type inferrer here

PR #2549, opened 2018-12-20
Add warning if out events are used but never raised

### LINKED ISSUE #2434: Add warning if out events are checked in reactions but are not raised before.
If out events are used as reaction triggers or in reaction guards but are not raised in the statechart then they will always evaluate to false. It makes sense to provide a hint to the user to explain the constraint and avoid confusion.

A validation rule with the following properties must be implemented. 
- [ ] The validation rule creates a warning with the message "Out event <event name> is never raised in this statechart and thus is never active".
- [ ] Check out events that are used as trigger or part of the guard expression on transitions and local reactions.
- [ ] The condition is true if there is no raise statement for the out event in the whole statechart. 

This issue was derived from issue #2205 .

tangele:
fixes #2434

### DIFF
--- a/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextValidator.xtend
+++ b/plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextValidator.xtend
@@ -316,15 +316,14 @@ class STextValidator extends AbstractSTextValidator implements STextValidationMe
 		
 		var boolean isRaised = false;
 		for (Setting setting : usages) {
-			if (!(setting.EObject instanceof Scope) && //
-				(setting.getEObject().eContainer instanceof EventRaisingExpression)) {
+			if (setting.getEObject().eContainer instanceof EventRaisingExpression) {
 				isRaised = true;
 			}
 		}
 		
 		if (!isRaised) {
 			for (Setting setting : usages) {
-				if (!(setting.EObject instanceof Scope) && (setting.EObject instanceof ElementReferenceExpression)) {
+				if (setting.EObject instanceof ElementReferenceExpression) {
 					warning(String.format(OUT_EVENT_NEVER_RAISED, event.getName()), setting.EObject, null, -1);
 				}
 			}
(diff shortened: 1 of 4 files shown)


test-plugins/org.yakindu.sct.model.stext.test/src/org/yakindu/sct/model/stext/test/validation/STextJavaValidatorTest.java
```diff
@@ -652,6 +652,19 @@ public void checkUnusedEntry() {
...
+	public void checkNotRaisedOutEvent() {
+		statechart = AbstractTestModelsUtil.loadStatechart(VALIDATION_TESTMODEL_DIR + "NeverRaisedOutEvent.sct");
+		Diagnostic diagnostics = Diagnostician.INSTANCE.validate(statechart);
+	
+		// used as transition trigger and inside reaction guard
+		assertWarning(diagnostics, String.format(OUT_EVENT_NEVER_RAISED, "B"));
+		
+		// used inside reaction guard
+		assertWarning(diagnostics, String.format(OUT_EVENT_NEVER_RAISED, "A"));
+		assertIssueCount(diagnostics, 4);
```
tkutz: You have 4 issues but only check two of them, is this fine or is there something missing?

plugins/org.yakindu.sct.model.stext/src/org/yakindu/sct/model/stext/validation/STextValidator.xtend
```diff
@@ -307,6 +309,23 @@ class STextValidator extends AbstractSTextValidator implements STextValidationMe
...
 		}
 	}
+	
+	@Check(CheckType.FAST)
+	def void checkNotRaisedOutEvent(EventDefinition event) {
+		if (event.direction != Direction.OUT) {
+			return;
+		}
+		var Collection<Setting> usages = EcoreUtil.UsageCrossReferencer.find(event, event.eResource().getResourceSet());
+		var boolean isRaised = usages.findFirst[setting | setting.getEObject().eContainer instanceof EventRaisingExpression] !== null;
```
BeckmaR: Nice usage of xtend. Even better suited for the task is [`exists`](http://download.eclipse.org/modeling/tmf/xtext/javadoc/2.4/org/eclipse/xtext/xbase/lib/IterableExtensions.html#exists(java.lang.Iterable,%20org.eclipse.xtext.xbase.lib.Functions.Function1)).

PR #2558, opened 2019-01-02
Fix clashes of C/C++ 'null' macro

### LINKED ISSUE #2555: Definition of 'null' in sc_types.h clashes with alternative definitions especially in C++
There are some problems with the definition of 'null' in sc_types.h in the context of C++:

- it does not make use of existing C++ definitions
- 'null' is likely subject to definition clashes. As an example Qt 5.10 upwards also defines that symbol.

Additionally the literal name is defined by a constant and thus is not easily changeable for platform specific code generators.

So the following actions must be taken
- [ ]  replace the 'null' by 'sc_null' 
- [ ] redefine sc_null properly in sc_types.h for C
- [ ] redefine sc_null sc_types.h for C++ to be 'NULL'
- [ ] replace constant by extension method in C & C++ code generators

terfloth:
fixes #2555

### DIFF
--- a/test-plugins/org.yakindu.sct.generator.cpp.test/model/test.sgen
+++ b/test-plugins/org.yakindu.sct.generator.cpp.test/model/test.sgen
@@ -322,34 +322,34 @@ GeneratorModel for sctunit::cpp {
 		}
 	} 
 	
-	test ConstantsVariationTests{
-		
-		feature Outlet{ 
-			targetProject = targetProject
-			targetFolder = "gtests/ConstantsVariationTest"
-			libraryTargetFolder = "libraryTarget"			
-		}		
-		
-		feature JUnitWrapper {
-			WrapToJUnit =  true
-		}			
-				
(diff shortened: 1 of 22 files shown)


test-plugins/org.yakindu.sct.generator.cpp.test/model/test.sgen
```diff
@@ -322,34 +322,34 @@ GeneratorModel for sctunit::cpp {
...
-			inlineEnterRegion = false
-			inlineExitRegion = false
-			inlineEntries = false
-		}
-		
-		feature SGenModel {
-			GenerateSGen =  true
-		}
-	}
+//	test ConstantsVariationTests{
```
andreasmuelder: Why is this test commented out? Can you fix it or remove it if it is not required anymore?

terfloth:
Follow to clarify status and pending actions of ConstantsVariationTests is defined at #2559.

PR #2654, opened 2019-02-06
Reworded some chapter beginnings …

RainerKlute:
… in order to start with a short paragraph that is more suitable for the tile navigation on the itemis website.


RainerKlute:
Is already live on the website: https://www.itemis.com/en/yakindu/state-machine/documentation/user-guide

PR #2656, opened 2019-02-06
Add "Install Dependencies.." button to example wizard

### LINKED ISSUE #2653: Add means to easily install dependencies of examples
In our example wizard we list a couple of example projects which have dependencies to other features, either coming from our update site, or from a third party update site.

Currently the user has to read the description text in manually install all dependencies in order to use an example. It would be more convenient if the wizard would allow to directly open the installation wizard with all the example's dependencies being preselected.

tkutz:
Fixes #2653 

Merge this one first: https://github.com/Yakindu/statecharts/pull/2652


andreasmuelder:
nice!

PR #2672, opened 2019-02-14
move arrays to default type system

andreasmuelder:


plugins/org.yakindu.base.types/src/org/yakindu/base/types/typesystem/GenericTypeSystem.xtend
```diff
@@ -0,0 +1,67 @@
...
+		declareSuperType(getType(ANY), getType(INTEGER))
+		declareSuperType(getType(INTEGER), getType(REAL))
+	}
+
+	protected def declareArrayType() {
+		val array = createArrayType()
+		array.abstract = true
+		declareType(array, array.getName())
+		getResource().contents += array
+	}
```
tkutz: `declareArrayType` is never called

PR #2673, opened 2019-02-14
Use 'f' suffix for FloatLiterals

### LINKED ISSUE #2391: C-Code Generator should use proper suffixes for values
I just found a case where there's actually a difference between `1.7` and `1.7f`, such that the comparison `1.7 == 1.7f` actually returns false.

BeckmaR:
Fixes #2391


andreasmuelder:
We should not add this to the base generator since other code generators (python, javascript) do not support literal suffixes.

Issue #2476, opened 2018-12-03
Quickfix or code completion for target project


rherrmannr:
Sometimes it's annoying to type in the exact targetProject name in the feature outlet.

A quickfix or code completion in the tragetProject string would be helpful.

In the sgen file:
``targetProject = "SomeProjectWithAVeryLongNameCanBePlacedHere"``

PR #2694, opened 2019-03-05
add quickfix to change targetProject

tangele:
fixes https://github.com/Yakindu/statecharts/issues/2476


plugins/org.yakindu.sct.generator.core/src/org/yakindu/sct/generator/core/library/AbstractDefaultFeatureValueProvider.java
```diff
@@ -35,6 +36,9 @@
 
 	protected static final SGenFactory factory = SGenFactory.eINSTANCE;
 	
+	public static final String ERROR_CODE_PROJECT = "code_error";
```
andreasmuelder: missing project?

PR #2720, opened 2019-03-26
Issue 2690

### LINKED ISSUE #2690: Transitions with multiple exit point connections are not handled correctly
For some reason, if a transition specifies multiple handled exit points like `# e1> e2>`, only the first one works correctly, whereas the other one breaks the statechart. See example:

image

When simulating this, either event `e1` or `e2` should have the same effect - **StateA** is left, `x` is set to 1, **1** is active. This only works for `e1`. `e2` deactivates **StateA**, does not activate **1**, and does not set `x` to 1.

This can also be observed in the generated code (everything inlined):

```

/* The reactions of exit e1. */
static void react_main_region_StateA__region0_e1(Default3* handle)
{
	/* The reactions of exit e1. */
	/* Default exit sequence for state StateA */
	/* Default exit sequence for region  */
	/* Handle exit of all possible states (of default3.main_region.StateA._region0) at position 0... */
	switch(handle->stateConfVector[ 0 ])
	{
		case Default3_main_region_StateA__region0_State1 :
		{
			/* Default exit sequence for state State1 */
			handle->stateConfVector[0] = Default3_last_state;
			handle->stateConfVectorPosition = 0;
			break;
		}
		default: break;
	}
	/* Exit action for state 'StateA'. */
	handle->iface.x = 1;
	/* 'default' enter sequence for state 1 */
	handle->stateConfVector[0] = Default3_main_region__1;
	handle->stateConfVectorPosition = 0;
}

/* The reactions of exit e2. */
static void react_main_region_StateA__region0_e2(Default3* handle)
{
}
```

While exit `e1` properly reacts, exit `e2` does nothing.

OTRS #2019022737000101

terfloth:
fixes #2690


plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/transformation/ReactionBuilder.xtend
```diff
@@ -146,6 +146,35 @@ class ReactionBuilder {
...
+	 */
+	def defaultExitTransition(State it) {
+		outgoingTransitions.filter( t | t.hasNoTrigger && t.explicitlyHandlesExit('default')).head
+	}
+	
+	/**
+	 * Returns the names of all exit points specified by a transition.
+	 */
+	def exitPoints(Transition it) {
+		 properties.filter(typeof(ExitPointSpec)).map( eps | eps.exitpoint).toList
```
BeckmaR: I may be wrong, but should the line not look like this?
```suggestion
		 properties.filter[ExitPointSpec].map[ eps | eps.exitpoint].toList
```
Especially the square brackets for `map`.

PR #2722, opened 2019-03-27
Fix NPE for declarations in internal scope

tkutz:


tkutz:
See https://github.com/Yakindu/statecharts/pull/2723

PR #2738, opened 2019-04-04
initial SText quickfix implementation for unknown events

tangele:
Also, also a serializer bug has been fixed, which only occoured while
testing the SText Quickfixes. Given a Definition-Section containing only
the default annotations, the quickfix for creating both the event and
its corresponding unnamned interface would eat up all spaces between the
Annotations and the newly created interface.

That caused syntactic failures, as the Annotation and the interface
could not be correctly parsed anymore - no whitespace was between the
keywords.

### DIFF
--- a/plugins/org.yakindu.sct.model.stext.ui/src/org/yakindu/sct/model/stext/ui/quickfix/EventTriggerCreationCommand.xtend
+++ b/plugins/org.yakindu.sct.model.stext.ui/src/org/yakindu/sct/model/stext/ui/quickfix/EventTriggerCreationCommand.xtend
@@ -1,0 +1,86 @@
+/** 
+ * Copyright (c) 2019 itemis AG and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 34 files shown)


plugins/org.yakindu.sct.model.stext.ui/src/org/yakindu/sct/model/stext/ui/quickfix/EventTriggerCreationCommand.xtend
```diff
@@ -0,0 +1,92 @@
...
+	protected EObject contextElement;
+	protected Issue issue;
+	Direction desiredDirection
+
+	new(EObject element, Issue issue, Direction desiredDirection) {
+		super(TransactionUtil.getEditingDomain(element), "Create missing event", Collections.emptyList);
+		if (issue.data.length < 2) {
+			throw new IllegalArgumentException(
+				"data[0] is expected to contain an URI from a embedded resource to the statechart, " +
+					"data[1] is expected to contain the semantic name of the invalid EObject"
```
tkutz: `issue.data is expected to have two entries: (1) an URI to the statechart resource and (2) the name of the unresolved reference.`

tkutz:
Looks good to me!

Only thing is that in some scenarios the newly created events get too many newlines before them, but that is an issue on its own.

PR #2744, opened 2019-05-06
Improve tree shortener fault tolerance

BeckmaR:
Catches a case where strings could be put twice into the string shortener, introducing a bug where all "`react"`-functions had the same name.

### DIFF
--- a/plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/naming/tree/TreeNamingService.xtend
+++ b/plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/naming/tree/TreeNamingService.xtend
@@ -51,19 +51,17 @@ class TreeNamingService implements INamingService {
 	@Inject extension ElementNameProvider
 
 	@Inject protected StringTreeNodeDepthComparator stringTreeNodeDepthComparator
-	
+
 	@Inject protected IStringShortener shortener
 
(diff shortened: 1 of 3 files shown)


plugins/org.yakindu.sct.model.sexec/src/org/yakindu/sct/model/sexec/naming/tree/TreeNamingService.xtend
```diff
@@ -345,35 +348,27 @@ class TreeNamingService implements INamingService {
...
-	
+
 	def protected List<String> addSeparator(List<String> segments) {
 		val List<String> result = newArrayList
-		for(var i = 0; i < segments.size(); i++) {
-			result.add(segments.get(i)) {
-				if(i < segments.size() - 1) {
+		for (var i = 0; i < segments.size(); i++) {
+			result.add(segments.get(i))
+			{
```
tkutz: This is not our usual formatting style

tkutz: Or in other words, you can just remove the brackets

BeckmaR: Yeah that's old code, I didn't reformat it all.

PR #2756, opened 2019-05-14
Set travis build distro to trusty

BeckmaR:
Precise has been long deprecated, everything works the same with trusty. Plus maven output with colors, so that's cool.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
-sudo: false
 language: java
-dist: precise
+dist: trusty
 jdk: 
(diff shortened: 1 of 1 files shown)


andreasmuelder:
This did prevent the migration so far: 
https://github.com/travis-ci/travis-ci/issues/8172

BeckmaR:
We could try and merge this, if the bug is still present in Travis we just revert.

andreasmuelder:
Did you try the dist: xenial ?

PR #2807, opened 2019-09-12
Fixed #2805

### LINKED ISSUE #2805: 'Space' in FileNames leads to problems
When creating a StatechartModel File with the eclipse wizard the name of the file can contain spaces.
These get translated to '%20', which then leads to an invalid identifier name in the StateChart View.

Either the Wizard and the StateChart name should be allowed containing those symbols or neither.

Mazesch:
added a check on file creation with the following regex `^[-a-zA-Z0-9._]+`

### DIFF
--- a/plugins/org.yakindu.sct.ui/src/org/yakindu/sct/ui/wizards/ModelCreationWizardPage.java
+++ b/plugins/org.yakindu.sct.ui/src/org/yakindu/sct/ui/wizards/ModelCreationWizardPage.java
@@ -70,6 +70,11 @@ public class ModelCreationWizardPage extends WizardNewFileCreationPage {
 			setErrorMessage(NLS.bind("file extension is not valid! The valid extension is: " + "." + fileExtension, extension));
 			return false;
 		}
+		String name = getFileName();
+		if (name != null && !name.matches("^[-a-zA-Z0-9._]+")) {
+			setErrorMessage("file name is not valid!");
+			return false;
+		}
 		return true;
 	}
 


plugins/org.yakindu.sct.ui/src/org/yakindu/sct/ui/wizards/ModelCreationWizardPage.java
```diff
@@ -70,6 +70,11 @@ protected boolean validatePage() {
 			setErrorMessage(NLS.bind("file extension is not valid! The valid extension is: " + "." + fileExtension, extension));
 			return false;
 		}
+		String name = getFileName();
+		if (name != null && !name.matches("^[-a-zA-Z0-9._]+")) {
+			setErrorMessage("file name is not valid!");
```
andreasmuelder: You can reuse the error msg from line 70

Mazesch: But the error that is catched in these lines has nothing to do with the extension, which could lead to users being confused, because they have the "`.sct`"-extension;
for example the name "`test file.sct`" which gets translated to "`test%20file.sct`" which is invalid not because of its extension but because it contains a space(or just an invalid name in generel)

tkutz: I agree with @Mazesch

PR #2812, opened 2019-09-19
open online example documentation when available, refactor

tangele:


plugins/org.yakindu.sct.examples.wizard/src/org/yakindu/sct/examples/wizard/service/ExampleModelOpener.java
```diff
@@ -38,26 +49,53 @@
...
+				}
+			} catch (PartInitException | MalformedURLException e) {
+				e.printStackTrace();
+				return;
+			}
+		});
+	}
+
+	protected void openExampleReadme(IProject project) throws PartInitException {
+		if (ExampleData.DESC_FILE != null) {
```
BeckmaR: Remove this if block - this is a `static final String` and not null.

PR #2822, opened 2019-10-10
Changed the codeblocks on the website

### LINKED ISSUE #2806: Code on the web tutorial does not work out of the box
In the [tutorial](https://www.itemis.com/en/yakindu/state-machine/documentation/tutorials/tutorial_code-generation#tutorial_code-generation) under the section 2.4.6 on the itemis page the code snippet  for the CallHandlingClient doesn't work out of the box, because:

- The imports don't work because they are located somewhere else

- The Timer doesn't work

- The CallHandlingStatemachine does not work out of the box

Mazesch:
The codeblocks now work with the examples as seen in the screenshots for [this](https://www.itemis.com/en/yakindu/state-machine/documentation/tutorials/tutorial_code-generation#tutorial_code-generation) tutorial
fix #2806

### DIFF
--- a/plugins/org.yakindu.sct.doc.user/src/tutorials/tutorials.textile
+++ b/plugins/org.yakindu.sct.doc.user/src/tutorials/tutorials.textile
@@ -417,17 +417,7 @@ We have to enter the respective definitions in textual form into the statechart 
 
 bc.. 
 @EventDriven
-// Use the event driven execution model.
-// Runs a run-to-completion step
-// each time an event is raised.
-// Switch to cycle based behavior
-// by specifying '@CycleBased(200)'
-// instead.
-
 @ChildFirstExecution
-// In composite states, execute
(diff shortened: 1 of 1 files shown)


plugins/org.yakindu.sct.doc.user/src/tutorials/tutorials.textile
```diff
@@ -415,7 +415,21 @@ In the _CallHandling_ example, we have two actors: the user and the phone. Let's
...
 
-bc.. interface User:
+bc.. 
+@EventDriven
+// Use the event driven execution model.
+// Runs a run-to-completion step
+// each time an event is raised.
+// Switch to cycle based behavior
+// by specifying '@CycleBased(200)'
+// instead.
```
BeckmaR: I think the comments can be removed

PR #2833, opened 2019-10-11
Added a function for issue#2774

Mazesch:
the addeed function checks if the give function(in the parameter) was called
exactly `times` times

this also requires a merged pullrequest in sct_pro


tkutz:
Closing and changing documentation instead

PR #2836, opened 2019-10-16
Apply style in non-scaling case also

### LINKED ISSUE #2835: State names are no more bold
Since version 3.4.0, state names are not bold anymore. At least I can see this effect on my macOS. As far as I know, this wasn't changed on purpose.

tkutz:
Fixes #2835


tkutz:
@andreasmuelder please check on windows

PR #2845, opened 2019-10-24
Embedded systems integration guide

rherrmannr:
Examples: https://github.com/Yakindu/examples/pull/161


plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
+
+h1(#embedded_systems_integration_guide). Embedded Systems Integration Guide
+
+State Machine code generated by YAKINDU Statechart Tools can effectively be used on Embedded Systems. The integration contains different parts, but can mostly be described in a generic way, as most steps of the integration are platform independent.
```
BeckmaR:
```suggestion
State machine code generated by YAKINDU Statechart Tools can effectively be used on embedded systems. The integration contains different parts, but can mostly be described in a generic way, as most steps of the integration are platform independent.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+The differences of the integration mostly stemp from the execution - CycleBased or EventDriven - of the state machine or the used functionalities, such as in, out and timed events. Other best practice implementation patterns can be based on whether a bare metal microcontroller or an operating system is used. So, there are different factors which may have a impact on the integration:
```
BeckmaR:
```suggestion
The differences of the integration mostly stem from the execution - CycleBased or EventDriven - of the state machine or the used functionalities, such as in, out and timed events. Other best practice implementation patterns can be based on whether a bare metal microcontroller or an operating system is used. So, there are different factors which may have an impact on the integration:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+* Bare-metal or RTOS
+* Usage of interrupts or polling
+* Usage of sleep modes
+
+This guide will give an overview of the most common and well known design patterns in a generic way and initially provides examples for an Arduino and only starts with the bare-metal integration. It will be continuously expanded.
+
+h2(#polling-vs-interrupts). Polling vs.Interrupts
+
+Programming bare-metal microcontrollers without any kind of RTOS or scheduler mainly leads to the architectural decision: Polling or Interrupts.
+This section will give a short overview, of how both design patterns work.
```
BeckmaR:
```suggestion
This section will give a short overview of how both design patterns work.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	while(true){
+		readInputs();
+		doStuff();
+		updateOutputs();
+		delay_ms(100);
+	}
+}
+
+p. 
+The program starts with an initialization. After this, everything will be handled in a while loop, where the inputs will be polled cyclic - in this case very 100 milliseconds plus the execution time the program. Wating for x milliseconds must not be realized in a blocking delay function, it also can be done by using sleep modes and a timer interrupt. But the main idea of polling is cyclic reacting on events.
```
BeckmaR:
```suggestion
The program starts with an initialization. After this, everything will be handled in a while loop, where the inputs will be polled cyclically - in this case every 100 milliseconds plus the execution time the program. Waiting for x milliseconds does not need to be realized in a blocking delay function, it also can be done by using sleep modes and a timer interrupt. But the main idea of polling is cyclic reacting on events.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+For applications where events/inputs only rarely occur, this is not a good design choice, as the sensor are frequently read. This will lead to higher power consumption.
```
BeckmaR:
```suggestion
For applications where events/inputs only rarely occur, this is not a good design choice, as the sensors are read frequently. This will lead to higher power consumption.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+h3(#interrupts). Interrupts
+
+Using interrupts is the second approach, where the system reacts on interrupts, which can occur asynchronously. Thus, the system must not cyclic poll the inputs. The normal execution of the program will be interrupted and it jumps to the Interrupt Service Routine (ISR), which then will be executed. Typically, the execution of the ISR should be as fast as possible, as other interrupts could occur while executing the ISR. Therefore, bool flags are commonly used, which serve as memory. A simple implementation look like this:
```
BeckmaR:
```suggestion
Using interrupts is the second approach, where the system reacts on interrupts, which can occur asynchronously. Thus, the system does not need to cyclically poll the inputs. The normal execution of the program will be interrupted and it jumps to the Interrupt Service Routine (ISR), which then will be executed. Typically, the execution of the ISR should be as fast as possible, as other interrupts could occur while executing the ISR. Therefore, bool flags are commonly used, which serve as memory. A simple implementation look like this:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+bc.. 
+int main() {
+	while(true) {
+		if(bool_flag1 == true) {
+			doStuff();
+			bool_flag1 = false;
+		}
+		if(bool_flag2 == true) {
+			doDiffferentStuff();
```
BeckmaR:
```suggestion
			doDifferentStuff();
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+ISR_2 {
+	bool_flag2 = true;
+}
+
+p. 
+Each ISR uses one specific boolean flag, which stores wherever an interrupt has occurred. This pattern is often combined with a sleep mode, so that the microcontroller's CPU is only active after an interrupt. This pattern is often used for low power applications, which are powered by batteries.
+
+h2(#Example-State-Machine-for-Integration). Example State Machine for Integration
+
+The complexity of the state machine can raise in a development process. Thus, it's recommended to specify of how the API (Applicable Programming Interface) should be accessed. While the API depends on the execution and used event types, this guide tries to break down the main elements and define a expandable design pattern. Therefore, the following example will be used for all integration guides.
```
BeckmaR:
```suggestion
The complexity of the state machine can increase in a development process. Thus, it's recommended to specify how the API (Application Programming Interface) should be accessed. While the API depends on the execution and used event types, this guide tries to break down the main elements and define am expandable design pattern. Therefore, the following example will be used for all integration guides.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+!(large)images/images_embedded_systems_integration_guide/platform_integration_example.jpeg(Example State Machine)!
+
+p=. Example State Machine
+
+The state machine contains three types of events: two __in events__ (inEvent1, inEvent2), two __out events__ (outEvent1, outEvent2) and one __timed event__ (after 2 s). Two states are connected with the inEvent1 and the after 2 s event. Each state raises one of __out event__. Additionally, a final state has been added, which also is part of the API. It only can be reached by raising the __in event__ inEvent2, if the state machine's __state__ StateA is active.
```
BeckmaR:
```suggestion
The state machine contains three types of events: two __in events__ (inEvent1, inEvent2), two __out events__ (outEvent1, outEvent2) and one __timed event__ (after 2 s). Two states are connected with the inEvent1 and the `after 2 s` event. Each state raises one of __out event__. Additionally, a final state has been added, which also is part of the API. It only can be reached by raising the __in event__ inEvent2, if the state machine's __state__ StateA is active.
```

Please check correct textile syntax

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+* While loop - calling the state machine until it's final
+* Final - called if the state machine is final
+The execution of the program can be visualized as a state machine:
+
+!(large)images/images_embedded_systems_integration_guide/platform_integration_simple.png(High level description of bare metal cycle based state machine integration)!
+
+p=. High level description of bare metal cycle based state machine integration
+
+
+ According to the used State Machine, there are different parts, which must be take into account:
```
BeckmaR:
```suggestion
 According to the used State Machine, there are different parts, which must be taken into account:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+* In Events 
+* Timed Events
+* Out Events
+
+h3(#main-function). Main Function
+
+The main function could be realized as following: First, the hardware will be initialized. Then the timers will take into account. After this the state machine will be initialized and entered.
```
BeckmaR:
```suggestion
The main function could be realized as following: First, the hardware will be initialized. Then the timers will be taken into account. After this the state machine will be initialized and entered. 
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+ According to the used State Machine, there are different parts, which must be take into account:
+* In Events 
+* Timed Events
+* Out Events
+
+h3(#main-function). Main Function
+
+The main function could be realized as following: First, the hardware will be initialized. Then the timers will take into account. After this the state machine will be initialized and entered. 
+At this point the state machine will be called as long as the state machine is not final. Note: If there is no final state defined the behavior is equals a while(true) loop. At first all in events will be raised. After this the timer will be updated, which internally raises time events. The behavior equals an in events. Then the runCycle will be called, which handles the logical execution of the state machine. This only needed, if the execution semantics is cycle based and must only be executed based on the elapsed time, e.g. 200 ms for a CycleBased(200) state machine. After proceeding, the out events will be handled. The last step, if wanted, is setting the microcontroller into a sleep mode.
```
BeckmaR:
```suggestion
At this point the state machine will be called as long as the state machine is not final. Note: If there is no final state defined the behavior is equals a while(true) loop. At first all in events will be raised. After this the timer will be updated, which internally raises time events. The behavior is similar to in events. Then the runCycle will be called, which handles the logical execution of the state machine. This is only needed if the execution semantics is cycle based and must only be executed based on the elapsed time, e.g. 200 ms for a CycleBased(200) state machine. After proceeding, the out events will be handled. The last step, if wanted, is setting the microcontroller into a sleep mode.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+			goToSleep()
+			// wait for Interrupt
+		}
+	}
+}
+p. 
+
+h3(#handle-in-events). Handle In Events 
+
+For now on, there are two different ways of how the events should be interact on the embedded system. One design pattern is polling. The other one is the usage of interrupts.
```
BeckmaR:
```suggestion
For now on, there are two different ways of how the events should interact on the embedded system. One design pattern is polling. The other one is the usage of interrupts.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+A design using polling cyclic updates the status of the inputs and raise the according event.
```
BeckmaR:
```suggestion
A design using polling cyclically updates the status of the inputs and raises the respective event.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+Sensor2_ISR{
+	inEvent2 = true;
+	inEvent2Value = readSensor2();
+}
+p. 
+
+h3(#handle-timer). Handle Timer
+
+Handling the timer is more comprehensive. Each time event of the state machine must be updated and handled, because in detail there are nothing else than a in event. This means that for every timed trigger there must be one in event. Therefore, a timer service is provided in the examples. This timer service must be updated with the elapsed time since the its last call, which leads to the next implementation point: Determining the time.
```
BeckmaR:
```suggestion
Handling the timer is more comprehensive. Each time event of the state machine must be updated and handled, because in detail they are nothing else than an in event. Therefore, a timer service is provided in the examples. This timer service must be updated with the elapsed time since its last call, which leads to the next implementation point: Determining the time.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+On a pure bare-metal implementation there is no such thing as elapsed time. Everything depends on cycles. 
+
+Imagine a microcontroller running with a clock of 16MHz. Running one instruction needs 1/16MHz = 625ns. A delay() function 1000000 would need 6,25 ms.A first approach updating the timer could be:
```
BeckmaR:
```suggestion
Imagine a microcontroller running with a clock of 16MHz. Running one instruction needs 1/16MHz = 62.5ns. A delay() function 100000 would need 6,25 ms. A first approach updating the timer could be:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+bc.. 
+void main() {
+	while(true) {
+		delay(1000000);
```
BeckmaR:
```suggestion
		delay(100000);
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+		updateTimer(6,25);
+		runCycle();
+	}
+}
+p. 
+
+But using this implementations got two big drawbacks:
+* the delay function is blocking
+* the execution time of the updateTimer and runCycle are not taking into account
```
BeckmaR:
```suggestion
* the execution time of the updateTimer and runCycle are not taken into account
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+A much more precise design is using timer interrupts, which can generate a periodic interrupt ever x cycles:
```
BeckmaR:
```suggestion
A much more precise design is using timer interrupts, which can generate a periodic interrupt every x cycles:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	}
+}
+
+#define TIMER_TICK_MS 32 // 32 ms for example
+Timer_ISR{ 
+	updateTimerFlag = true;
+}
+p. 
+
+Some microcontrollers, like the Arduino, support functions like millis(), which are in fact using the a similar method, but storing the elapsed time since starting the device. This time can be used to determine the elapsed time between two cycles in a loop:
```
BeckmaR:
```suggestion
Some microcontrollers, like the Arduino, support functions like millis(), which are in fact using a similar method, but storing the elapsed time since starting the device. This time can be used to determine the elapsed time between two cycles in a loop:
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+bc.. 
+void loop() {
+	current_millies = millis();
```
BeckmaR:
```suggestion
	current_millis = millis();
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	updateTimer(current_millies - last_cycle_time);
```
BeckmaR:
```suggestion
	updateTimer(current_millis - last_cycle_time);
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	statemachine_runCycle();
+	last_cycle_time = current_millies;
```
BeckmaR:
```suggestion
	last_cycle_time = current_millis;
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+}
+
+p. 
+
+For prototyping, this implementation can be used, but it should be considered that the timer will overflow after approximately 50 days. That's why we highly recommend a implementation with interrupts as described before.
+
+h3(#handle-out-events). Handle Out Events
+
+Out events are raised by the state machine can be checked by using the state machine's API. If an out event gets raised its returning value is true, as long as the next runCycle is be executed. This is why the out events should be handled <b>after</b> every runCycle call. They can be used to wire up different actuators. Mapping values, e.g. integers, to the out events, is also possible.
```
BeckmaR:
```suggestion
Out events raised by the state machine can be checked by using the state machine's API. If an out event gets raised its returning value is true, until the next runCycle is executed. This is why the out events should be handled <b>after</b> every runCycle call. They can be used to wire up different actuators. Mapping values, e.g. integers, to the out events, is also possible.
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
+
+h1(#embedded_systems_integration_guide). Embedded Systems Integration Guide
+
+State machine code generated by YAKINDU Statechart Tools can effectively be used on embedded systems. The integration contains different parts, but can mostly be described in a generic way, as most steps of the integration are platform independent.
+The differences of the integration mostly stem from the selected execution semantics - cycle based or event driven - as well as the used functionalities, such as in, out and timed events. Other best practice implementation patterns can be based on whether a bare metal microcontroller or an operating system is used. So, there are different factors which may have an impact on the integration:
```
terfloth: perhaps shortly describe the purpose of this chapter ...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+h2(#Bare-metal-cycle-based-integration). Bare-Metal Integration
+
+This section describes how to integrate a state machine with a cycle based or event driven execution semantic on any microcontroller, so on any embedded device. The integration can be structured in three parts, which results the main part of your program.
+* Initialization - called once
+* While loop - calling the state machine until it's final
+* Final - called if the state machine is final
+The execution of the program can be visualized as a state machine:
+
+!(large)images/images_embedded_systems_integration_guide/platform_integration_simple.png(High level description of bare metal cycle based state machine integration)!
+
```
terfloth: I won't use a meta state machine to describe the code structure here. It could be confusing to introduce two state machines on two different meta levels. The overall structure already became clear from the coding pattern described above.

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	if(hasTimeEvents) {
+		timerInit();
+	}
+	
+	statemachine_init();
+	statemachine_enter();
+
+	// While loop
+	while(!statemachine_isFinal()) {
+		if(hasInEvents) {
```
terfloth:
```suggestion
		#if (hasInEvents)
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+			handleInEvents();
+		}
```
terfloth:
```suggestion
		#endif
```

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
```
terfloth: We mix meta levels if we define it as a runtime expression. If we want to describe it in C-style then we shoudl use preprocessor directives.

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+	// While loop
+	while(!statemachine_isFinal()) {
+		if(hasInEvents) {
+			handleInEvents();
+		}
+		
+		if(hasTimeEvents) {
+			handleTimer();
+		}
```
terfloth: see above...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+		
+		if(isCycleBased && elapsedTime >= cyclePeriod) {
+			elapsedTime = 0;
+			statemachine_runCycle();
+		}
```
terfloth: see above...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+		
+		if(hasOutEvents) {
+			handleOutEvents();
+		}
```
terfloth: see above...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+		
+		if(usingSleepMode) {
+			goToSleep()
+			// wait for Interrupt
+		}
```
terfloth: see above...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
```diff
@@ -0,0 +1,281 @@
...
+
+h3(#execution-as-state-machine). Execution As State Machine
+
+This is a expanded view on the implementation, of how it can be realized as a state machine.
+(needs to be updated)!
+
+!(large)images/images_embedded_systems_integration_guide/platform_integration.png(Complete execution shown as State Machine)!
+
+p=. Complete execution shown as State Machine
+
```
terfloth: it is already described and what is describe is a meta state machine. I won't go for too much meta here ...

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
rherrmannr: uuuuupps :D

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
rherrmannr: :+1:

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
rherrmannr: :+1:

plugins/org.yakindu.sct.doc.user/src/user-guide/embedded_systems_integration_guide.textile
rherrmannr: :+1:

PR #2849, opened 2019-10-25
fixed #2774

### LINKED ISSUE #2774: SCTUnit assert operation call not working as documented
Hello,
I'm testing a state machine using operation call assertion and I noticed that if I do

assert called sStopPwm [x] times

it returns me true whenever sStopPwm operation is called a numbert of time >= [x] instead from your documentation it should assert an ==.
"The third assertion checks whether the operation myOperation has been called exactly 4 times"
best regards
Michele

Mazesch:
changed the word `exactly` to `atleast`, because that is the behavior

### DIFF
--- a/plugins/org.yakindu.sct.doc.user/src/user-guide/sctunit.textile
+++ b/plugins/org.yakindu.sct.doc.user/src/user-guide/sctunit.textile
@@ -440,7 +440,7 @@ The first assertion checks whether the operation _myOperation_ has been called d
 
 The second assertion not only checks whether the operation _myOperation_ has been called, but also checks whether it has been called with parameters 42 and 815. If the operation hasn't been called at all, the test fails. The test also fails if the operation has been called, but with different parameters than 42 and 815, e.g., @myOperation(1, 2)@.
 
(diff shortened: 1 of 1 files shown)


plugins/org.yakindu.sct.doc.user/src/user-guide/sctunit.textile
```diff
@@ -440,7 +440,7 @@ The first assertion checks whether the operation _myOperation_ has been called d
 
 The second assertion not only checks whether the operation _myOperation_ has been called, but also checks whether it has been called with parameters 42 and 815. If the operation hasn't been called at all, the test fails. The test also fails if the operation has been called, but with different parameters than 42 and 815, e.g., @myOperation(1, 2)@.
 
-The third assertion checks whether the operation _myOperation_ has been called exactly 4 times, no matter the arguments, while the fourth assertion checks if the operation has been called exactly one time with the specified parameters.
+The third assertion checks whether the operation _myOperation_ has been called atleast 4 times, no matter the arguments, while the fourth assertion checks if the operation has been called exactly one time with the specified parameters.
```
tkutz: atleast => at least

PR #2868, opened 2019-11-08
Internal & In Event Queue size

### LINKED ISSUE #2702: C code generator: rework allocation of in and internal event queues
The in event queue as implemented for #2617 applies the same allocation schema as the internal event queue. The user must be able to apply different dimensions for internal and in event queue. Additionally different instances of the state machine may require different in event queue sizes depending on the usage context. So we have to rework the allocation for in event queues.

- [ ] decouple in and internal event queue dimensions
- [x] make in event queue size configurable per instance
- [x] rename struct type x_internal_event to x_event

### LINKED ISSUE #2729: Document inEventQueue option for C code generator

BeckmaR:
Makes the size of the internal and in-event queue configurable, but the user has to provide a buffer to the init function. This is configurable with a new sgen-option `userAllocatedQueue`.

Fix #2702
Fix #2729

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/library/FeatureTypeLibrary.xmi
+++ b/plugins/org.yakindu.sct.generator.c/library/FeatureTypeLibrary.xmi
@@ -50,7 +50,12 @@
       optional="true">
     <parameters
         name="inEventQueue"
+        comment="If set to true, a queue for incoming events is generated."
         optional="true"
+        parameterType="BOOLEAN"/>
+    <parameters
+        name="userAllocatedQueue"
+        comment="If set to true, the buffers for the event queues are not allocated by the state machine, but have to be provided to the init function instead."
         parameterType="BOOLEAN"/>
   </types>
 </sgen:FeatureTypeLibrary>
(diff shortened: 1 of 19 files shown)


plugins/org.yakindu.sct.generator.c/library/FeatureTypeLibrary.xmi
```diff
@@ -50,7 +50,12 @@
       optional="true">
     <parameters
         name="inEventQueue"
+        comment="If set to true, a queue for incoming events is generated."
         optional="true"
         parameterType="BOOLEAN"/>
+    <parameters
```
terfloth: Must be optional

test-plugins/org.yakindu.sct.generator.c.test/gtests/InEventQueueTest/InEventQueueTest.cc
```diff
@@ -37,7 +37,9 @@ void inEventQueueIface_askOracle(InEventQueue* handle) {
 
 
 TEST(InEventQueue, InEventQueue) {
-	inEventQueue_init(&machine);
+	ineventqueue_event internal_queue[20];
```
terfloth: This tests : in queue / internal queue / user allocation
There should be additional tests:
* in queue / internal queue / without user allocation
* in queue / without user allocation
* internal queue / without user allocation
* in queue / with user allocation
* internal queue / with user allocation

PR #2904, opened 2020-01-10
Init call for multi sm

rherrmannr:


rherrmannr:
should not be part of sequencer

PR #2910, opened 2020-01-14
Refactor C Multi SM generator by using OriginTracing

rherrmannr:


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/Naming.xtend
```diff
@@ -397,72 +379,43 @@ class Naming {
...
+	def dispatch access(EObject it) '''#error cannot access elements of type «getClass.name»'''
 
-	def valueAccess(Declaration it){
-		val container = eContainer
-		if(container instanceof ComplexType){
-			return '''iface.«name.asIdentifier.value»'''
-		}
-		'''«scHandle»->«scope.instance».«name.asIdentifier.value»'''
-	}
+	def dispatch valueAccess(Declaration it)'''«IF needsHandle»«scHandle»->«ENDIF»«scope.instance».«name.asIdentifier.value»'''
```
tkutz: What happens if `scope` is null, e.g. because the declaration comes from a header type (C domain)?

PR #2919, opened 2020-01-16
remove duplicate bundle from feature

Bananeweizen:
Branding plugin was contained twice in the feature (see line 46).

Signed-off-by: Michael Keppler <michael.keppler@gmx.de>


andreasmuelder:
Right, thanks!

PR #2969, opened 2020-02-06
Multi sm c op calls

rherrmannr:
`submachine.opCall`!

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/CExpressionsGenerator.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/CExpressionsGenerator.xtend
@@ -176,7 +176,7 @@ class CExpressionsGenerator extends ExpressionsGenerator {
 	}
 
 	def dispatch CharSequence code(FeatureCall it,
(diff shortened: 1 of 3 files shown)


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/CExpressionsGenerator.xtend
```diff
@@ -217,4 +214,20 @@ class CExpressionsGenerator extends ExpressionsGenerator {
...
+	protected def CharSequence argCode(List<Expression> expressions)
+		'''«FOR arg : expressions BEFORE ', ' SEPARATOR ', '»«arg.code»«ENDFOR»'''
+	
+	protected def CharSequence argCode(ElementReferenceExpression it) {
+		expressions.argCode
+	}
+	
+	protected def CharSequence argCode(FeatureCall it) {
+		expressions.argCode
+	}
```
tkutz: You can put all `argCode` methods together into:
```
protected def CharSequence argCode(ArgumentExpression it)
		'''«FOR arg : expressions BEFORE ', ' SEPARATOR ', '»«arg.code»«ENDFOR»'''
```

rherrmannr: you're totally right

PR #3040, opened 2020-03-13
Split up initialization for ifaces to avoid in warnings

### LINKED ISSUE #3039: [Cpp Compiler] Using 'this' in initialization list causes warning C4355
The compiler warning:
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4355?view=vs-2019

This is only the case for **EventDriven** statecharts, as they hold an instance of the parent state machine. **CycleBased** statecharts do not need the ``this`` argument.
```
Default::Default()  :
stateConfVectorPosition(0),
iface(this)
{
}
```

The problem: 

> You should not use the this pointer until all construction has completed.

I would suggest the following fix for **EventDriven** state machines:
```
Default::Default()  :
stateConfVectorPosition(0),
iface(sc_null)
{
	iface = this;
}
```

rherrmannr:
fix #3039


rherrmannr:
``iface = this`` uses the implicit constructor. I will change this and update the initialization list for interfaces.

PR #3062, opened 2020-04-09
Use sc_uinteger instead of int for observer_count

rherrmannr:


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/files/RxCHeader.xtend
```diff
@@ -27,23 +27,14 @@ class RxCHeader implements IContentTemplate {
...
-		 *
-		 * Created on: 30.03.2020
-		 *     Author: terfloth
-		 */
 		«entry.licenseText»
 		
 		#ifndef «rxcModule.define»_H_
 		#define «rxcModule.define»_H_
 		
+		#include "stddef.h"
```
terfloth, line 35: please remove include - when we use sc_uinteger instead of size_t

rherrmannr: :+1:

PR #3077, opened 2020-05-04
C forward declarations

rherrmannr:


rherrmannr:
will be solved with https://github.com/Yakindu/statecharts/pull/3080

PR #3084, opened 2020-05-07
Explicitly init variables of complex types with null literal

### LINKED ISSUE #3083: Explicitly init variables of complex types with null literal
Currently, statechart variables of a complex type, e.g. a submachine, are not initialized at all in the init method of the statechart. We should explicitly set them to null, as different target languages might treat uninitialized variables differently.

This is also the reason why some multi-statemachine tests for the C++ code generator currently fail.

tkutz:
Fixes #3083


rherrmannr:
This will not work with CDomain.
```
struct_t myStruct = sc_null; //compile error!
```

need to filter multiSM types.

tkutz:
@rherrmannr Your comment should be fixed now.

rherrmannr:
@tkutz for testing you can execute the CDomain generator tests

rherrmannr:
I've tested this & it's working. LGTM...

PR #3085, opened 2020-05-11
C getters feature

rherrmannr:
Getters feature allows using old API. Observable are the new default.
```
feature OutEventAPI {
	observables = true // Default: true
	getters = true // Default: false
}
```
Extendable for out event callbacks.

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/GeneratorPredicate.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/GeneratorPredicate.xtend
@@ -28,4 +28,12 @@ class GeneratorPredicate extends org.yakindu.sct.generator.core.extensions.Gener
 	def useOutEventObservables() {
 		entries.getOutEventObservablesUsed(entry)
 	}
+	
+	def useOutEventGetters() {
+		entries.getOutEventGettersUsed(entry)
+	}
+	
+	def needsClearOutEventsFunction() {
+		!useOutEventObservables || useOutEventGetters
+	}
 }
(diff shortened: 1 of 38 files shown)


plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/Naming.xtend
```diff
@@ -241,12 +241,14 @@ class Naming {
...
-		if(useOutEventObservables && direction == Direction.OUT) {
-			return '''«OBSERVABLE_TYPE» «eventName»;'''
-		}
 		'''
-		«BOOL_TYPE» «eventRaisedFlag»;
-		«IF type !== null && type.name != 'void'»«typeSpecifier.targetLanguageName» «eventValueVariable»;«ENDIF»
+		«IF useOutEventObservables && direction == Direction.OUT»
+			«OBSERVABLE_TYPE» «eventName»;
+		«ENDIF»
+		«IF (useOutEventObservables && direction == Direction.OUT) || direction == Direction.IN || direction == Direction.LOCAL»
```
terfloth, line 248: must be:  IF (useOutEventGetters ...
currently _raised members are only generated if observables is active. Thus code cannot be compiled if setting is: observables = false getter = true

plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/GeneratorPredicate.xtend
```diff
@@ -28,4 +28,12 @@ class GeneratorPredicate extends org.yakindu.sct.generator.core.extensions.Gener
 	def useOutEventObservables() {
 		entries.getOutEventObservablesUsed(entry)
 	}
+	
+	def useOutEventGetters() {
+		entries.getOutEventGettersUsed(entry)
+	}
+	
+	def needsClearOutEventsFunction() {
+		!useOutEventObservables || useOutEventGetters
```
terfloth, line 37: simply: return useOutEventGetters 
Observables have no impact on getters. If we require that getters should be used if seen specifies "observables = false" and defines nothing for getters  then we should implement this in useOutEventGetters accordingly.

rherrmannr:
retest this please

PR #3087, opened 2020-05-12
[3.x] Handle is not required using C Header declarations

rherrmannr:
https://github.com/Yakindu/statecharts/pull/2986


rherrmannr:
already merged

Issue #3071, opened 2020-04-22
Deadlock using @SuperStep with inEventQueue test


rherrmannr:
1. open inEventQueue.sct
2. add @SuperStep
3. raise event ``f`` at least once
4. raise event ``e``
-> Deadlock

tkutz:
There is a cycle in the statechart, so the interpreter runs forever:
![Bildschirmfoto 2020-05-13 um 11 48 44](https://user-images.githubusercontent.com/2589669/81797963-bb70c780-950f-11ea-90be-de386fcc798e.png)

Cycle detection is hard as we could also have guard expressions involved, like this one in state L:
![Bildschirmfoto 2020-05-13 um 11 49 46](https://user-images.githubusercontent.com/2589669/81798066-dfcca400-950f-11ea-9084-9031b36af285.png)

But at least we should make the simulation cancelable, and not run in the UI thread directly.

PR #3089, opened 2020-05-13
Issue 3071 super step cycles

tkutz:
This one does not detect the cycle as stated in issue #3071 but it gives the user a chance to terminate the simulation, and it highlights the micro steps in a super step execution.

### DIFF
--- a/plugins/org.yakindu.sct.simulation.core.sexec/src/org/yakindu/sct/simulation/core/sexec/interpreter/DefaultExecutionFlowInterpreter.xtend
+++ b/plugins/org.yakindu.sct.simulation.core.sexec/src/org/yakindu/sct/simulation/core/sexec/interpreter/DefaultExecutionFlowInterpreter.xtend
@@ -197,8 +197,6 @@ class DefaultExecutionFlowInterpreter implements IExecutionFlowInterpreter, IEve
 
 				traceInterpreter.evaluate(endRunCycleTrace, executionContext)
 			} while (event !== null)
-		} catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


tkutz:
SuperSteps1

tkutz:
Checked that the thread gets destroyed when terminating a simulation, and also checked that there are not multiple threads created when wildly clicking on events in the view.

plugins/org.yakindu.sct.simulation.core.sexec/src/org/yakindu/sct/simulation/core/sexec/interpreter/DefaultExecutionFlowInterpreter.xtend
```diff
@@ -205,11 +207,13 @@ class DefaultExecutionFlowInterpreter implements IExecutionFlowInterpreter, IEve
...
 		}
 	}
-
+	
 	def superStepLoop(()=>void microStep) {
 		do {
 			stateVectorChanged = false
 			microStep.apply
+			// give the highlighter a chance to do its work
+			Thread.sleep(70)
```
terfloth, line 216: We can not sleep 70ms within each superstep cycle !!

tkutz: Replaced Thread.sleep by an explicit check for Thread.isInterrupted(). Now the user can click on terminate, but it might still take some seconds until it is performed.

PR #3091, opened 2020-05-14
Remove MaxPermSize vm argument

Bananeweizen:
This VM argument is not supported anymore with Java 8 and above. Since
the BREE of all Yakindu SCT bundles is 8, the argument is not needed
anymore in the plugins.

Also remove it from all development related artifacts (test launches,
eclipse.ini in oomph setup), since developers should use a Java 8+
runtime, too.

The most visible difference of this change is: no more JVM warnings when
setting up a new Yakindu SCT Oomph setup (when generating all the model
code via launch configs).


andreasmuelder:
thanks for the contribution, looks good to me.

PR #3100, opened 2020-05-29
Fix "Editor already locked" exception when running multiple simulatio…

tkutz:
…ns on same statechart

Since releasing and locking runs asynchronously, they need to be invoked in same async runnable to have a determined execution order.


plugins/org.yakindu.base.gmf.runtime/src/org/yakindu/base/gmf/runtime/highlighting/HighlightingSupportAdapter.java
```diff
@@ -195,12 +190,7 @@ public synchronized void releaseEditor() {
 			throw new IllegalStateException("Editor not locked!");
 		}
 		List<Action> singletonList = new ArrayList<>();
-		singletonList.add(new Action() {
-			@Override
-			public void execute(IHighlightingSupport hs) {
-				releaseInternal();
-			}
-		});
+		singletonList.add((support) -> releaseInternal());
```
terfloth, line 193: Name method 'releaseEditorInternal' instead of 'releaseInternal' for consistency with 'lockEditorInternal'?

tkutz: Done.

PR #3104, opened 2020-06-04
Travis CI: Use lowest target to which we are compatible

tkutz:
Currently, we build with 2019-12.target on Travis for the PR builds, but later on, the master will be built with Oxygen.target. And this step can fail when we use constructs that are not available in Oxygen.


andreasmuelder:
+1

PR #3107, opened 2020-06-05
Move OutEventAPI feature to core

rherrmannr:


plugins/org.yakindu.sct.generator.core/src/org/yakindu/sct/generator/core/library/ICoreLibraryConstants.java
```diff
@@ -41,5 +41,9 @@
 	String FUNCTION_INLINING_FEATURE_INLINE_ENTRIES = "inlineEntries";
 	String FUNCTION_INLINING_FEATURE_INLINE_ENTER_REGION = "inlineEnterRegion";
 	String FUNCTION_INLINING_FEATURE_INLINE_EXIT_REGION = "inlineExitRegion";
+	
+	public static final String FEATURE_OUT_EVENT_API = "OutEventAPI";
+	public static final String PARAMETER_OUT_EVENT_OBSERVABLES = "observables";
+	public static final String PARAMETER_OUT_EVENT_GETTERS = "getters";
```
tkutz, line 47: no need for `public static final` in interface (this is the default)

PR #3131, opened 2020-06-23
replace fornax generator

andreasmuelder:


andreasmuelder:
builds hangs to some random bugs, we stay with fornax for now

Issue #2260, opened 2018-09-18
Add Marquee Tool to Palette for multi selection of elements 


andreasmuelder:
It is very hard to multi-select nested elements. Users wish to have a selection tool similar to the marquee tools from papyrus: wiki.eclipse.org/Papyrus_User_Guide/Palette/Marquee_Selection_Documentation

PR #3132, opened 2020-06-24
use marquee selection as default drag tool within regions

### LINKED ISSUE #2995: Feature Request: Editor: Enable multi selection with mouse drag inside regions
Currently you can not select multiple statements with mouse drag, instead you move the region. It would be nice to have a special selection tool or fix the position of the regions.

andreasmuelder:
fixes #2995


andreasmuelder:
also fixes https://github.com/Yakindu/statecharts/issues/2260

PR #3144, opened 2020-06-30
C++ event buffer

### LINKED ISSUE #3139: Add event buffer for cycle based C++ state machines
Implement #3113 for C++ code generator. See #3114

rherrmannr:
fix #3139


plugins/org.yakindu.sct.generator.cpp/src/org/yakindu/sct/generator/cpp/CppKeywords.java
```diff
@@ -13,6 +13,5 @@
 
 public final class CppKeywords {
 	
-	public static final String[] CPP_KEYWORDS = { "clearInEvents", "clearOutEvents", "enter",
-		"exit", "runCycle", "parent"};
+	public static final String[] CPP_KEYWORDS = {"parent"};
```
tkutz, line 16: Why are enter, exit, runcycle no more marked as keywords? The methods will still be generated.

rherrmannr: They are now part of the Sequencer. The NamingService detects them and changes the ``runCycle`` method to ``runCycle_ID``.

PR #3147, opened 2020-07-01
[3.x] Always use iface name for operation access

### LINKED ISSUE #3146: Iface names are missing using feature IdentifierSettings
By using the IdentifierSettings feature, the iface names disappear in the generated C code. Operations are affected.

Define an interface like this:
```
interface named:
operation someOp()
```
results the following prototype in the required header:
```
extern void myXsmIfaceNamed_someOp(const MyXsm* handle);
```
Turning on the IdentifierSettings feature like this:
```
GeneratorModel for yakindu::c {

	statechart myXsm {

		feature Outlet {
			targetProject = "test"
			targetFolder = "src-gen"
			libraryTargetFolder = "src"
		}
		
		feature IdentifierSettings {
			moduleName = "MyXsm_SM"
			statemachinePrefix = "MyXsm_SM"
			separator = "_"
		}
	}
}
```
results this function prototype:
```
extern void MyXsm_SM_someOp(const MyXsm_SM* handle);
```
where the name of the iface ``named`` is missing.

However, by defining an operation with the same name in another unnamed iface like this:
```
interface:
operation someOp()
interface named:
operation someOp()
```
results:
```
extern void myXsmIface_someOp(const MyXsm* handle);

extern void myXsmIfaceNamed_someOp(const MyXsm* handle);
```

The generated functions should "stable" and not change by adding operations in different interfaces.

rherrmannr:
fix #3146

### DIFF
--- a/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/EventNaming.xtend
+++ b/plugins/org.yakindu.sct.generator.c/src/org/yakindu/sct/generator/c/extensions/EventNaming.xtend
@@ -27,7 +27,7 @@ class EventNaming {
 	@Inject extension INamingService
 
 	def eventEnumMemberName(Event it) {
-		'''«scope.functionPrefix(it)»_«name.asIdentifier»'''
+		'''«scope.functionPrefix»_«name.asIdentifier»'''
 	}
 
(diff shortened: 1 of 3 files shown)


rherrmannr:
this will only be available in v4, as it is API breaking.

PR #3148, opened 2020-07-01
Always use iface name for operation access

### LINKED ISSUE #3146: Iface names are missing using feature IdentifierSettings

rherrmannr:
fix #3146

### DIFF
--- a/test-plugins/org.yakindu.sct.generator.c.test/gtests/InternalEventLifeCycle_Naming/InternalEventLifeCycleTest.cc
+++ b/test-plugins/org.yakindu.sct.generator.c.test/gtests/InternalEventLifeCycle_Naming/InternalEventLifeCycleTest.cc
@@ -32,14 +32,14 @@ TEST_F(InternalEventLifeCycleTest, InternalEventLifeCycleTest) {
 	intlEvent_enter(&statechart);
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r1_A));
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r2_C));
-	intlEvent_raise_e(&statechart);
+	intlEvent_Iface_raise_e(&statechart);
 	sc_timer_service_proceed_cycles(&timer_service, 1);
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r1_A));
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r2_D));
 	sc_timer_service_proceed_cycles(&timer_service, 1);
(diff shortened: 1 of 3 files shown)


test-plugins/org.yakindu.sct.generator.c.test/gtests/InternalEventLifeCycle_Naming/InternalEventLifeCycleTest.cc
```diff
@@ -32,14 +32,14 @@ TEST_F(InternalEventLifeCycleTest, InternalEventLifeCycleTest) {
 	intlEvent_enter(&statechart);
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r1_A));
 	EXPECT_TRUE(intlEvent_isStateActive(&statechart, intlEvent_r2_C));
-	intlEvent_raise_e(&statechart);
+	intlEvent_Iface_raise_e(&statechart);
```
terfloth, line 35: Default naming is 'stateMachineIface_raise_e'.  The prefix (bad name) should simply substitute the state machine name.  So the function name should be 'intlEventIFace_raise_e' if the prefix 'intlEvent' is specified. The user can specify 'intEvent_' if he wants 'intlEvent_Iface_raisee'. So do not add the separator '_' by default.

rherrmannr: Yes :+1:

